PR #162, opened 2015-09-23
Tests : make gtest path relative to the parent of the PROJECT_SOURCE_DIR

aeantipov:
not the CMAKE_SOURCE_DIR (otherwise failure if module is compiled separately)


egull:
makes sense.

PR #186, opened 2016-02-04
Update lmod module file to proper structure

ryanlevy:
Now consistent with [tutorial](https://github.com/ALPSCore/ALPSCore/wiki/Tutorial%3A-link).


aeantipov:
Thanks! In homebrew they also require to squash commits in one, but we're not that formal here )

Issue #257, opened 2016-10-21
More tests for the linear, quadratic, logarithmic meshes are welcome


galexv:
Transplanted from issue #250 (specifically https://github.com/ALPSCore/ALPSCore/issues/250#issuecomment-252307129)

* It would be nice to have some tests that verify that linear, quadratic, and logarithmic meshes are indeed linear, quadratic and logarithmic. (I did not check the logic of their construction thoroughly; it may be easier to check the logic of the tests);
* Just a suggestion: what about moving the whatever_grid classes into a new grid namespace, to refer to them as grid::whatever ?
* It would be also nice to have a tutorial-like example on using the meshes and grids.

egull:
...for next time we go through the meshes...

PR #266, opened 2017-01-03
Real frequency mesh

iskakoff:
#257 + index_type for real frequency mesh


galexv:
Sergey (@iskakoff) --- thanks!

Issue #281, opened 2017-01-30
Using Legendre basis in gf


shinaoka:
Is someone already working on the implementation of Legendre basis for gf?

egull:
@iskakoff are you either using it or thinking of doing something?

shinaoka:
If no one has a concrete plan, I will implement this functionality myself for ALPSCore! Besides, I am also interested in implementing a container based on an expansion of G(tau) in terms of a "numerical" orthonormal basis set on the imaginary-time axis.

egull:
sounds good to me. Please go ahead – and keep us in the loop, so we can use your developments!

shinaoka:
OK. I will keep you updated. Actually, a private version of our CT-HYB code replies on a numerical basis set rather than the Legendre basis.

shinaoka:
It's done.

PR #291, opened 2017-02-16
Implemented Legendre basis

shinaoka:
Implemented Legendre representation of gf

### DIFF
--- a/gf/include/alps/gf/mesh.hpp
+++ b/gf/include/alps/gf/mesh.hpp
@@ -701,9 +701,11 @@ namespace alps {
             // It's a member function to avoid dealing with templated friend decalration.
             void swap(legendre_mesh& other) {
                 using std::swap;
+                if (this->statistics_ != other.statistics_) {
+                    throw std::runtime_error("Attempt to swap LEGENDRE meshes with different statistics");
+                }
                 swap(this->beta_, other.beta_);
                 swap(this->n_max_, other.n_max_);
-                swap(this->statistics_, other.statistics_);//SHOULD WE ALLOW TO SWAP MESHES WITH DIFFERENT STATISTICS?
                 base_mesh::swap(other);
             }
 
@@ -713,7 +715,6 @@ namespace alps {
                 ar[path+"/N"] << n_max_;
                 ar[path+"/statistics"] << int(statistics_);
                 ar[path+"/beta"] << beta_;
-                ar[path+"/points"] << points();
             }
 
             void load(alps::hdf5::archive& ar, const std::string& path)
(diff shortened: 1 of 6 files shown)


shinaoka:
This is about the issue #281.

egull:
@iskakoff please have a look at it. @galexv please have a look at it and merge it once @iskakoff OKs it.

iskakoff:
According to H5GF specification for Legendre mesh it should have the following structure:
 
    \+--kind:string="LEGENDRE"
    \---N :int[] # number of legendre points
    \---beta: double[] #inverse temperature
    \---statistics :int[] # 0:Bosonic 1:Fermionic

In your implementation you additionally store `points`. Is it really necessary? Because in `load` method you even don't use it and also in the `compute_points` it is just simple index of current point:

    _points().resize(extent());
    for(int i=0;i<n_max_;++i){
        _points()[i]=i;
    }

And if it's necessary, we should add optional `points` field to the H5GF specification.

You mentioned in the comment about swapping of two meshes with different statistic. Emanuel suggested to throw an exception when two meshes with different statistics are going to be swaped. I will do the same in Matsubara mesh, now It is a bug #292.

The rest is fine for me.

shinaoka:
Thank you for the review. Yes, you're right. I've modified the pull request. It now throws an exception when one attempts to swap two legendre meshes with different statistics. The points are not store in a HDF5 file.

gf/include/alps/gf/mesh.hpp
```diff
@@ -659,7 +659,124 @@ namespace alps {
...
+                int stat;
+
+                ar[path+"/N"] >> n_max;
+                ar[path+"/statistics"] >> stat;
+                ar[path+"/beta"] >> beta;
+
+                statistics_=statistics::statistics_type(stat);
+                beta_=beta;
+                n_max_=n_max;
+                check_range();
```
galexv: If check_range() throws (due to invalid HDF5 data), we end up with an object in an invalid state. 
It may not be a serious problem in practice, but it is probably better to read into a temporary object and then `swap()`.

iskakoff: @galexv actually we have the same check for Matsubara gf

galexv: Sure. :)  This is not the only place where we do not offer a strong exception guarantee. :(

But this can be addressed later and everywhere at once.

As i said, it is probably not a big deal in practice, so i am going to accept the merge.

Hiroshi (@shinaoka) and Sergey (@iskakoff), thanks for the contribution and the review!

PR #297, opened 2017-03-12
Implemented numerical mesh and piecewise polynomial

shinaoka:
I have cleaned up the code bit. More comments for Doxygen. It throws in the constructor if the object has not been constructed properly.

### DIFF
--- a/gf/include/alps/gf/mesh.hpp
+++ b/gf/include/alps/gf/mesh.hpp
@@ -9,17 +9,20 @@
 #include <boost/lexical_cast.hpp>
 #include <boost/operators.hpp>
 #include <boost/type_traits/integral_constant.hpp>
+#include <boost/type_traits/is_same.hpp>
 
 #include <alps/hdf5/archive.hpp>
 #include <alps/hdf5/complex.hpp>
 #include <alps/hdf5/vector.hpp>
 #include <alps/hdf5/multi_array.hpp>
 
+
 #ifdef ALPS_HAVE_MPI
 #include "mpi_bcast.hpp"
 #endif
 
 #include"flagcheck.hpp"
+#include"piecewise_polynomial.hpp"
 
 namespace alps {
     namespace gf {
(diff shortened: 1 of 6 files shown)


gf/include/alps/gf/mesh.hpp
```diff
@@ -779,6 +782,205 @@ namespace alps {
...
+            void save(alps::hdf5::archive& ar, const std::string& path) const
+            {
+                check_validity();
+                ar[path+"/kind"] << "NUMERICAL_MESH";
+                ar[path+"/N"] << dim_;
+                ar[path+"/k"] << k;
+                if (boost::is_same<T,double>::value) {
+                    ar[path+"/scalar_type"] << "double";
+                } else if (boost::is_same<T,std::complex<double> >::value) {
+                    ar[path+"/scalar_type"] << "complex_double";
```
galexv: If `T` is `std::complex<double>`, then the saved coefficients of the basis functions will anyway have "`__complex__`" attribute set, similar to the one described in https://github.com/ALPSCore/H5GF/blob/master/H5GF.rst#data-dataset . Do we need an extra "`scalar_type`" dataset here? 
Is there a situation when the `scalar_type` cannot be deduced from the type of `/basis_functions0/coefficients` dataset?
Or we want to increase data redundancy to make it more robust?

gf/include/alps/gf/mesh.hpp
```diff
@@ -779,6 +782,205 @@ namespace alps {
...
+        public:
+            typedef generic_index<numerical_mesh> index_type;
+
+            numerical_mesh(gf::statistics::statistics_type statistics=statistics::FERMIONIC):
+                    beta_(0.0), dim_(0), statistics_(statistics), valid_(false) {}
+
+            numerical_mesh(double b,  const std::vector<piecewise_polynomial<T,k> >&basis_functions,
+                           gf::statistics::statistics_type statistics=statistics::FERMIONIC):
+                    beta_(b), dim_(basis_functions.size()), basis_functions_(basis_functions), statistics_(statistics), valid_(false) {
+                set_validity();
```
galexv: > Hiroshi Shinaoka:
> I have cleaned up the code bit. More comments for Doxygen. 
> It throws in the constructor if the object has not been constructed properly.

@shinaoka : am i looking at the right code?
I do not see a `throw` in the constructor, somehow...
And the commit (95b1b5face01b0d5294555fadf87b6ca83d4e232) is said to be made 15 days ago...

shinaoka:
Oh. I made a mistake. I close this request and make a new one. Regarding your comment, scalar_type is not necessary.

PR #300, opened 2017-03-15
GFs and meshes are now default-constructive.

shinaoka:
Could you review the changes?
I have also introduced empty-gf checking.
It throws when an empty gf is accessed.
The overhead should be small.


galexv:
Hiroshi, my changes on top of these commits are delayed, so i'd better just merge and apply my changes later, when they will be ready.

PR #303, opened 2017-04-02
Removed template parameter "k" (order of polynomials) from

shinaoka:
I've changed the design of numerical mesh. This is for better compatibility with Python interfaces.


galexv:
Tests `MeshTest.MpiBcastRealspace` and `MeshTest.MpiBcastMomentum` (from `mesh_tes_mpi.cpp`) fail; apparently, because the receiving mesh (on a slave process) is empty and trips `throw_if_empty()` check.

I am looking at a possible solution now.

PR #336, opened 2017-07-05
Made define_convenience_parameters an inline function

shinaoka:
If my understanding is correct, this function should be inlined.


aeantipov:
Agree.

egull:
Thanks!

PR #353, opened 2017-10-06
Eigen as dependency

galexv:


CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ include(ALPSCompilerVersionCheck)
 include(ALPSCompilerTweaks)
 
 include(ALPSEnableMPI)
+include(ALPSEnableEigen)
```
mwallerb: Isn't the CMake convention not usually: `Find<Packagename>.cmake`? Why the deviation here?

common/cmake/ALPSCommonModuleDefinitions.cmake
```diff
@@ -233,6 +233,15 @@ endmacro(gen_pkg_config)
...
 # Function: generates main ALPSCore config
 function(gen_cfg_main)
+  # message("DEBUG: Eigen version is ${ALPS_HAVE_EIGEN_VERSION}")
+  # if (TARGET eigen)
+  #   message("DEBUG: OK, Eigen target is global")
+  #   get_target_property(prop_ eigen INTERFACE_INCLUDE_DIRECTORIES)
+  #   message("DEBUG: Eigen interface directories: ${prop_}")
+  # else()
+  #   message("DEBUG: No Eigen target is visible???")
+  # endif()
```
mwallerb: Do we need this commented code?

common/cmake/ALPSEnableEigen.cmake
```diff
@@ -0,0 +1,112 @@
+# enable using Eigen
```
mwallerb: This file is directly copied from the eigen distro, right?  In this case we should make a note to that effect in order to satisfy the BSD license . . .

mwallerb: Sorry, found it. Please ignore.

iskakoff:
Agree with @mwallerb , it will be a pain to keep track on changes in Eigen, especially the fact that we need Tensors that are in unsupported part of Eigen.
Why don't just use it as a find_package?

CMakeLists.txt
iskakoff: because it's bundled I suppose. Find<PackageName>.cmake is for external dependency.

krivenko:
Not that I'm going to argue too much, but I generally agree with @mwallerb.

Bundling Eigen to the source codes of ALPSCore has serious implications.
By doing so you

* agree to maintain our own copy of Eigen in a reasonable manner;
* make application writers use a very specific version of the library (they are not supposed to mix different versions in one project, are they?). What happens if the bundled version is buggy and broken for some particular use? Updating/downgrading the bundled sources is not a good solution, as they can easily be broken in some other place.

I think it sometimes makes sense to bundle small, old and/or poorly supported libraries. Eigen barely fits into this category.

galexv:
I did not use it as a `find_package` because there might not be a package available in the user's system. The conclusion of preliminary discussions was that we are to bundle Eigen with ALPSCore rather than encumber a user with requirements to install additional dependencies.  However, i agree that we do not want to maintain Eigen -- so we will just download it automatically at configure time and install it as part of ALPSCore "make install"  *unless* it is already available, in which case we treat it as we treat, e.g., Boost or MPI.

mwallerb:
Since I was the one that originally suggested bundling Eigen, I feel I should jump in and defend @galexv's honour here :)  The whole thing is AFAIAC a case of semantic confusion . . . what I meant with bundling was the following, three-step process in a way:

 1. a clean git repository (without dependencies)
 2. developer/maintainer checks out git, calls `cmake .. -DCMAKE_BUILD_TYPE=Bundle` or something, which generates a release tarball `alpscore-****.tgz` with the dependencies bundled in, which is then published
 3. user downloads release tarball, unpacks, and builds it.

Now, Alex informed me that step 2 is performed automatically by the github infrastructure, so it is not an option.  The next best thing, however is to download the thing with cmake, since that still allows the user to create a bundle and copy to some cluster.

galexv:
Sorry, there seems to be some mix-up: because i pushed new changed to the same branch that has an outstanding PR, the PR incorporates 2 unrelated groups of changes:

1. Versioning of ALPSCore;
2. Dependency on Eigen.

I am going to merge the versioning changes and rename this PR.

galexv:
If there is no Eigen in the system, which one to download?

1) The latest stable version, identified by its version number? I do not see any dedicated "latest" download URL. 
    Then as Eigen is updated and we want to depend on the new version, the requested version number in our CMake script has to change.

2) The tip of Eigen's Mercurial tree? 
    Then (a) Mercurial (`hg`) becomes a build dependency; (b) The tip of the tree is presumably unstable.

3) The latest "version" tag from the Mercurial tree? (https://bitbucket.org/eigen/eigen/downloads/?tab=tags)
   Then (a) Mercurial (`hg`) becomes a build dependency; (b) Some additional logic needed to determine the latest version tag, and still it can pull some rc version rather than a stable one.

My suggestion is to download `http://bitbucket.org/eigen/eigen/get/${EIGEN_VERSION}.tar.gz` (with a user-settable `EIGEN_VERSION` with a current default of `3.3.4`). If there is a special requirement of a newer version, it can always be downloaded manually by an ALPSCore user, and ALPSCore recompiled with the downloaded version.

egull:
I vote for the stable (3.3.4 is good). I would also require eigen after 3.2.0, as this is when the tensors were introduced. I checked that the release contain the unsupported headers (and in fact the distributions seem to have them too).

shinaoka:
Personally, I am using the very latest unstable version
because clang 3.9 fails to build the tensor module in 3.3.4.
The bug in the tensor module has been fixed after the release of 3.3.4.
It would be nice if the requirement is updated once 3.3.5 is released.

galexv:
Thanks for all the comments.
I am closing this PR unmerged, and i am going to remove the `devel` branch and branch the new one from the master, and make a new PR against the new one.

PR #354, opened 2017-10-12
Eigen as dependency

galexv:
This is an update/rewrite of PR #353 .

ALPSCore code can now use Eigen3 (`add_eigen()` CMake function provided by `ALPSEnableEigen.cmake` file).

1. If CMake  variable `ALPS_INSTALL_EIGEN` is `false` (the default), ALPSCore uses whatever Eigen is available on the local system. It tries to find it using (modified) `FindEigen3.cmake` --- either a CMake-based installation or a directory pointed to by `EIGEN3_INCLUDE_DIR` CMake var. A relevant variable is `ALPS_EIGEN_MIN_VERSION` (current default is "3.3.4") --- the minimal version required by ALPSCore.  Usage:
```sh
$ cmake -DEIGEN3_INCLUDE_DIR=/path/to/installed/eigen/headers [-DALPS_EIGEN_MIN_VERSION=3.3.4] <other arguments>
```

Note that in this case Eigen3 must remain in the same directory (presumably, system-wide installation directory) at the time of linking of client's code with ALPSCore. 

2.  If CMake variable `ALPS_INSTALL_EIGEN` is true, ALPSCore will co-install Eigen with ALPSCore. Eigen will be taken from the directory pointed to by `EIGEN3_INCLUDE_DIR` CMake var. If the variable is empty, ALPSCore will attempt to download Eigen3 (unless already downloaded) and unpack it. Upon `make install` Eigen will be copied from the source directory to ALPSCore installation tree.
Relevant variables:
  * `EIGEN3_INCLUDE_DIR` : if set, points to where to find Eigen to install from;
  * `ALPS_EIGEN_MIN_VERSION` : version of Eigen to download (default 3.3.4)
  * `ALPS_EIGEN_DOWNLOAD_LOCATION` : where to download Eigen from (default http://bitbucket.org/eigen/eigen/get/${ALPS_EIGEN_MIN_VERSION}.tar.gz)
  * `ALPS_EIGEN_UNPACK_DIR` : can be set if for some reason unpacking in the build directory won't do;
  * `ALPS_EIGEN_TGZ_FILE` : if set, specifies location of a pre-downloaded Eigen `*.tar.gz` archive.


shinaoka:
Hi, I am wondering how a client code can find the location of Eigen3.

There is a stupid question.
My current test is about the case of ALPS_INSTALL_EIGEN=False.
Should the client code invoke FindEigen3 in its own CMakeLists.txt?

As follows, some of the installed files of ALPSCore contain the location of Eigen3.
What is the intended way to use this information from a client's code?
> cd /opt/ALPSCore
./share/alps-accumulators/alps-accumulators.cmake:  INTERFACE_INCLUDE_DIRECTORIES "/opt/boost_1_63/include;/opt/local/include;/opt/Eigen3/include/eigen3;${_IMPORT_PREFIX}/include"
./share/alps-gf/alps-gf.cmake:  INTERFACE_INCLUDE_DIRECTORIES "/opt/boost_1_63/include;/opt/local/include;/opt/Eigen3/include/eigen3;${_IMPORT_PREFIX}/include"

galexv:
@egull @shinaoka There is an example at [tutorials/use_eigen/](https://github.com/ALPSCore/ALPSCore/tree/705c24aed8f33f03f65bf3c92733d2ef0460ff19/tutorials/use_eigen) . The situation is the same as with MPI dependency: the dependencies are "transitive". If you use ALPSCore module that uses Eigen, you have [`ALPSCore_HAS_EIGEN_VERSION` set after `find_package(ALPSCore)`](https://github.com/ALPSCore/ALPSCore/blob/705c24aed8f33f03f65bf3c92733d2ef0460ff19/tutorials/use_eigen/CMakeLists.txt#L15), and [adding ALPSCore dependency to your target](https://github.com/ALPSCore/ALPSCore/blob/705c24aed8f33f03f65bf3c92733d2ef0460ff19/tutorials/use_eigen/CMakeLists.txt#L26) automatically adds [Eigen headers](https://github.com/ALPSCore/ALPSCore/blob/705c24aed8f33f03f65bf3c92733d2ef0460ff19/tutorials/use_eigen/use_eigen.cpp#L10) to your include path.

egull:
Hm. I did that, and it didn't work. Possible that I have to rebuild alps from scratch to make it propagate through all the files.

galexv:
I believe you may need to remove the installed ALPSCore, and probably CMakeCache.txt, too. The only files it should propagate through are the generated *.cmake files in install/share/* .

egull:
yeah, not removing the CMakeCache was the problem. It works now.

galexv:
Hiroshi (@shinaoka) and Markus (@mwallerb), does this approach work for you?

shinaoka:
Yes, it works now!

PR #358, opened 2017-10-18
 Fixed crashes when running tests in parallel

mwallerb:
Some GF and accumulator tests request the same HDF5 filename,
which is a problem when running the tests in parallel using
`ctest -jXX`.  This assigns a unique filename to all files in
the test.

Also, it tightens some of the timeouts in the MC class.

### DIFF
--- a/mc/test/timer.cpp
+++ b/mc/test/timer.cpp
@@ -19,7 +19,7 @@ template <typename C>
 class TestCallback : public ::testing::Test {
   public:
     typedef C callback_type;
-    static const std::size_t TIMELIMIT=3; 
+    static const std::size_t TIMELIMIT=1;
     alps::params param_;
 
     TestCallback() {}
--- a/accumulators/test/custom_type.cpp
+++ b/accumulators/test/custom_type.cpp
@@ -173,7 +173,7 @@ struct CustomTypeAccumulatorTest : public testing::Test {
     }
 
     void TestSaveAccumulator() {
-        const std::string fname="save_acc.h5";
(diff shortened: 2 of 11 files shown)


mc/test/timer.cpp
```diff
@@ -19,7 +19,7 @@ template <typename C>
 class TestCallback : public ::testing::Test {
   public:
     typedef C callback_type;
-    static const std::size_t TIMELIMIT=3; 
+    static const std::size_t TIMELIMIT=1;
```
galexv: POSIX timer resolution is 1 sec. I am not sure that running simulation for 1 second is good enough test. Maybe 2 seconds? (and, does reducing test time by 2 seconds noticeably affects the total test run time?)

mwallerb:
On Wednesday, October 18, 2017 9:59:23 AM EDT Alexander Gaenko wrote:
> galexv commented on this pull request.
> 
> > @@ -19,7 +19,7 @@ template <typename C>
> 
>  class TestCallback : public ::testing::Test {
>    public:
>      typedef C callback_type;
> -    static const std::size_t TIMELIMIT=3;
> +    static const std::size_t TIMELIMIT=1;
> 
> POSIX timer resolution is 1 sec. I am not sure that running simulation for 1
> second is good enough test. Maybe 2 seconds? (and, does reducing test time
> by 2 seconds noticeably affects the total test run time?)

Yes.  It cuts the test time from 8 s to 2 s.

Also, I fail to see how the 2s case is fundamentally different from the 1s one

PR #361, opened 2017-10-31
[params] Correction to 92e92ea2b9a6807487d014c6f2bd107d49da50c6

krivenko:
`params_apply_visitor` should contain specialization of the call operator for `detail::trigger_tag`. Otherwise, user-provided functor `f` will be instantiated for this type (implementation detail leak).

### DIFF
--- a/params/include/alps/params/params_impl.hpp
+++ b/params/include/alps/params/params_impl.hpp
@@ -222,7 +222,7 @@ namespace alps {
             /// Even though this overload must be defined, it should never be called
             void operator()(const boost::optional<detail::trigger_tag>& val) const
             {
-              assert(false);
+                throw std::runtime_error("Internal error");
             }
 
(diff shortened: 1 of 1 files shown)


params/include/alps/params/params_impl.hpp
```diff
@@ -219,6 +219,12 @@ namespace alps {
                 throw option_type::visitor_none_used("Attempt to use uninitialized option value");
             }
 
+            /// Even though this overload must be defined, it should never be called
+            void operator()(const boost::optional<detail::trigger_tag>& val) const
+            {
+              assert(false);
```
egull: propose to throw rather than assert so it doesn't survive the debug mode, either.

krivenko: You mean the release mode, right?

egull: yes, sorry.

PR #379, opened 2017-11-15
Fixes most `-Wall` warnings

### LINKED ISSUE #375: Does not compile cleanly with `-Wall`
ALPSCore emits a huge strings of warnings with `-Wall`, particularly in the HDF5 and the accumulators modules.

This makes it very ugly to enable `-Wall` by default (ideally also `-Wextra`, but I'm not that `-pedantic`).  By disabling `-Wall`, however, we hide a lot of potential problems in the development stage.  Also, it is just not good style to have warning-ridden code.

mwallerb:
Fixes #375 by fixing most instances of two types of warnings:

 * signed/unsigned integer comparison
 * unused parameters

### DIFF
--- a/accumulators/include/alps/accumulators/feature/binning_analysis.hpp
+++ b/accumulators/include/alps/accumulators/feature/binning_analysis.hpp
@@ -255,7 +255,7 @@ namespace alps {
                         using alps::numeric::check_size;
 
                         B::operator()(val);
-                        if(static_cast<size_t>(B::count()) == (1UL << m_ac_sum2.size())) {
+                        if(B::count() == (1UL << m_ac_sum2.size())) {
                             m_ac_sum2.push_back(T());
                             check_size(m_ac_sum2.back(), val);
                             m_ac_sum.push_back(T());
--- a/accumulators/include/alps/accumulators/feature/binning_analysis.hpp
+++ b/accumulators/include/alps/accumulators/feature/binning_analysis.hpp
@@ -255,7 +255,7 @@ namespace alps {
                         using alps::numeric::check_size;
 
                         B::operator()(val);
-                        if(B::count() == (1 << m_ac_sum2.size())) {
+                        if(static_cast<size_t>(B::count()) == (1UL << m_ac_sum2.size())) {
                             m_ac_sum2.push_back(T());
                             check_size(m_ac_sum2.back(), val);
                             m_ac_sum.push_back(T());
--- a/accumulators/include/alps/accumulators/feature/binning_analysis.hpp
+++ b/accumulators/include/alps/accumulators/feature/binning_analysis.hpp
@@ -101,7 +101,7 @@ namespace alps {
             template<typename A> typename boost::disable_if<
                   typename has_feature<A, binning_analysis_tag>::type
                 , typename autocorrelation_type<A>::type
-            >::type autocorrelation_impl(A const & acc) {
+            >::type autocorrelation_impl(A const & /*acc*/) {
                 throw std::runtime_error(std::string(typeid(A).name()) + " has no autocorrelation-method" + ALPS_STACKTRACE);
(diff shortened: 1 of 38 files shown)


accumulators/include/alps/accumulators/feature/binning_analysis.hpp
```diff
@@ -255,7 +255,7 @@ namespace alps {
                         using alps::numeric::check_size;
 
                         B::operator()(val);
-                        if(B::count() == (1 << m_ac_sum2.size())) {
+                        if(static_cast<size_t>(B::count()) == (1UL << m_ac_sum2.size())) {
```
galexv: Do we need this `static_cast<size_t>`?  `count()` is supposed to return `count_type<T>::type` which is `boost::uint64_t`. Maybe it's better to redefine it in [count.hpp:31](https://github.com/ALPSCore/ALPSCore/blob/master/accumulators/include/alps/accumulators/feature/count.hpp#L31)?

accumulators/include/alps/accumulators/feature/max_num_binning.hpp
```diff
@@ -345,7 +345,7 @@ namespace alps {
                 void partition_bins(alps::mpi::communicator const & comm,
                                     std::vector<typename mean_type<B>::type> & local_bins,
                                     std::vector<typename mean_type<B>::type> & merged_bins,
-                                    int root) const
+                                    int /*root*/) const
```
galexv: `partition_bins()` does not use `root` argument and it is not needed for any overload -- shall we eliminate it? (I guess it is here as a placeholder, if one day we decide to use `MPI_Reduce()` rather than `MPI_Allreduce()`...)

mwallerb: Is it worth it? I mean the writing's on the wall anyway for `alps::accumulators`

mc/include/alps/mc/mpiadapter.hpp
```diff
@@ -72,15 +72,15 @@ namespace alps {
                             alps::mpi::all_reduce(communicator,
                                                   has_count,
                                                   std::plus<size_t>());
-                    if (sum_counts == communicator.size()) {
+                    if ((int)sum_counts == communicator.size()) {
```
galexv: Not a big deal, but why C-style cast `(int)` rather than `static_cast<int>()`, for consistency with other places?

mc/include/alps/mc/mpiadapter.hpp
```diff
@@ -72,15 +72,15 @@ namespace alps {
...
                         if (communicator.rank() == 0) {
                             typename Base::observable_collection_type::value_type merged = this->measurements[*it];
                             merged.collective_merge(communicator, 0);
                             partial_results.insert(*it, merged.result());
                         } else {
                             this->measurements[*it].collective_merge(communicator, 0);
                         }
-                    } else if (sum_counts > 0 && sum_counts < communicator.size()) {
+                    } else if (sum_counts > 0 && (int)sum_counts < communicator.size()) {
```
galexv: The same: not a big deal, but why C-style cast `(int)` rather than `static_cast<int>()`, for consistency with other places?

mc/include/alps/mc/mpiadapter.hpp
mwallerb: Lazyness, mainly. Will fix it.

utilities/test/test_utils_test.cpp
```diff
@@ -45,11 +45,11 @@ class TestUtilsTest : public ::testing::Test {
         const unsigned int nvals=range<value_type>::VALUE;
         std::vector<value_type> vals(nvals);
 
-        for (int i=0; i<nvals; ++i) {
+        for (int i=0; i<(int)nvals; ++i) {
```
galexv: Maybe better to change index `i` to unsigned for consistency...

Issue #295, opened 2017-03-06
collective_merge duplicates measurements


jgreitemann:
This is probably not a bug but a matter of design. Calling `accumulator_set::collective_merge(comm, 0)` will currently collect the measurements from all MPI processes into the master, but also leave them in the slave processes. Calling `collective_merge` _again_ would thus give wrong results as the slave measurements would be duplicated but the master measurements wouldn't. I'd consider this a design flaw.

In a typical Monte Carlo simulation, this is not a problem since `collect_results` is only ever called once at the end of the simulation. However, we found this to be an issue when doing bold diagrammatic Monte Carlo where the results are collected occasionally to improve the bold approximation self-consistently but the accumulators are intentionally not reset but continue to accumulate data.

As a simple workaround, one can save the measurements to a checkpoint prior to the collective_merge and restore the master from that checkpoint after the merge. One can also keep a copy of the `accumulator_wrapper`s in memory (obtainable using the `clone` function, but not the copy ctor! -- Why?). Unfortunately, the latter has to be done for each accumulator separately; one cannot just copy the `accumulator_set` as a whole since its copy ctor doesn't do anything either. (Maybe this could also be addressed? In the code, it says, _TODO: how do we handle that?_ I fail to see why it couldn't just deep-copy the `wrapper_set`.)

The above workarounds do however copy the data unnecessarily. As an alternative, one could just call `accumulator_wrapper::reset()` on the slaves after a collective merge. I would argue that this should not be left up to the user, but should in fact be done in `collective_merge`.

This [example program](https://gist.github.com/jgreitemann/e9649178d90f78b07498b250c1364d06) demonstrates the issue.

egull:
@galexv, can you give a status update? I just heard from Lode/Jonas that this one is critical for them.

galexv:
@egull Ok, it has just went to my top-priority task list.

mwallerb:
Just for mental bug-tracking for me: this is fixed in ALEA by commit d5c6c1e4 .
It is of course *not fixed* in the accumulators library.

galexv:
So, the current version of the fix (commit as of 9bd2ff2) does not merge the autocorrelation lengths properly. But should it?

[...]

PR #380, opened 2017-11-15
Resetting slave accumulators

galexv:
This is to fix issue #295: duplication on measurements by repeated collective_merge.
Autocorrelation values after repeated merge are not quite correct.

### DIFF
--- a/accumulators/include/alps/accumulators/accumulator.hpp
+++ b/accumulators/include/alps/accumulators/accumulator.hpp
@@ -871,10 +871,11 @@ namespace alps {
             public:
                 inline void collective_merge(alps::mpi::communicator const & comm, int root) {
                     boost::apply_visitor(collective_merge_visitor(comm, root), m_variant);
+                    if (comm.rank()!=root) this->reset();
                 }
-                inline void collective_merge(alps::mpi::communicator const & comm, int root) const {
-                    boost::apply_visitor(collective_merge_visitor(comm, root), m_variant);
-                }
+                // inline void collective_merge(alps::mpi::communicator const & comm, int root) const {
+                //     boost::apply_visitor(collective_merge_visitor(comm, root), m_variant);
+                // }
 #endif
 
             private:
--- a/accumulators/test/CMakeLists.txt
+++ b/accumulators/test/CMakeLists.txt
@@ -33,7 +33,8 @@ set (test_src
 if(ALPS_HAVE_MPI)
   set (test_src_mpi
     mpi_merge
-    mpi_merge_uneven    
+    mpi_merge_uneven
+    repeated_merge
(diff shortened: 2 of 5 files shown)


egull:
ok, looks good to me. But as this is in response to @jgreitemann 's report I will solicit a review from him.

jgreitemann:
Thanks for keeping me in the loop. The modifications look fine to me, however when running our code against it without the previously used workaround, I get some weird deadlock behavior when measuring a reset accumulator again. I suspect that's due to a problem on our end, but would like to understand what exactly is going on before greenlighting this PR. I'll probably have some time to investigate over the weekend.

galexv:
@jgreitemann 
> I get some weird deadlock behavior when measuring a reset accumulator again

I have had a similar experience, so let me venture a guess that might help figuring out what could have happened:
The collective merge now clears the accumulators on slaves. If there is no data added to accumulators prior to the next collective merge, the merge detects empty accumulator on slaves and throws an exception. If the exception propagates to the top level, `alps::mpi::environment` calls `MPI_Abort()`.  If the exception is caught (for example, by GTest framework), this results in unmatched collective MPI call and MPI waiting forever, which looks like a deadlock.

jgreitemann:
Thanks Alex, that is exactly what's happening. You saved me some pain there.

So, the exceptions stems from [this boost assert](https://github.com/ALPSCore/ALPSCore/blob/008bef5aaa93df92aff14a748b75e8c1c59cf3a6/accumulators/include/alps/accumulators/feature/binning_analysis.hpp#L267). I haven't been able to consistently reproduce it; there seems to be some delicate timing involved. In particular, I got the code to throw only when compiled with optimizations (which made debugging tricky), and from what I observed, it was always the master (rank 0) process that threw upon the first new measurement of any observable after some collective merges. I don't entirely understand the binning analysis code, but my best guess is that the `m_ac_partial` vector is not affected by the merge and that leads to problems when the binning level of the merged-in accumulator was higher than that of the accumulator on the master? Possibly the solution is as simple as adding something like `m_ac_partial.resize(m_ac_sum2.size());` after [line 356](https://github.com/ALPSCore/ALPSCore/blob/master/accumulators/include/alps/accumulators/feature/binning_analysis.hpp#L356)?

Strictly speaking, this probably isn't related to this PR, it's only that the specific circumstances to produce this error didn't show up before. Why I don't get the same problem with my original workaround (where I called `wrapper_set::reset()` on the `measurements` member in `run` *after* `collect_results` is done) escapes me.

egull:
Alex, could you please get this one done, too? Is there anything else holding it up?

galexv:
I have to do some debugging to reproduce the behavior observed by Jonas, and fix it...

PR #395, opened 2017-11-27
Output message for found wrong version of eigen3

ryanlevy:
When I was compiling `ALPSCore` my installation found my Eigen3 installation but continued to output the same message (which made me think I was doing something wrong). To help with the confusion the "header" of the message will now explain the wrong version has been found. An example:
```
-- eigen requested
-- Eigen3 version 3.3.1 found in /Users/ryle/modules/eigen331, but at least version 3.3.4 is required
-- Could NOT find Eigen3 (missing:  EIGEN3_VERSION_OK) (Required is at least version "3.3.4")
CMake Error at common/cmake/ALPSEnableEigen.cmake:52 (message):

The required library Eigen3 has been found at /Users/ryle/modules/eigen331 on your system.
   We require 3.3.4 > 3.3.1
   Please try upgrading or using the following options to use a different installation:
   1. Set environment variable Eigen3_DIR
...
```

### DIFF
--- a/common/cmake/ALPSEnableEigen.cmake
+++ b/common/cmake/ALPSEnableEigen.cmake
@@ -22,11 +22,38 @@ function(add_eigen)
 
   if (NOT ALPS_INSTALL_EIGEN)
     find_package(Eigen3 ${ALPS_EIGEN_MIN_VERSION})
-    if (NOT EIGEN3_FOUND) # CMake 3.3+ would use Eigen3_FOUND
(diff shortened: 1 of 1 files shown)


galexv:
Committed via rebase as 7119ea6d95cd0687025bc89dfda99f9606f3a859 and merged as of commit c9009b52c73b78371bb5b0e0851eab330e4b60a6. 
Thanks @ryanlevy!

PR #425, opened 2018-01-15
Fix paths in lmod file

ryanlevy:
`ALPSCore_DIR` needs to be the install/share/ALPSCore directory; `CMAKE_MODULE_PATH ` is for `FindALPScore.cmake` which isn't provided. 

Using just `ALPSCore_DIR=$root` leads to an error of  "Could NOT find ALPSCore (missing: ALPSCore_LIBRARIES)"

### DIFF
--- a/common/misc/alpscore.lmod.in
+++ b/common/misc/alpscore.lmod.in
@@ -3,7 +3,7 @@
 set     name        alpscore
 set     root        "@CMAKE_INSTALL_PREFIX@"
 
-setenv      ALPSCore_DIR    $root
+setenv      ALPSCore_DIR    $root/share/ALPSCore
+setenv      ALPSCore_ROOT   $root
 
 prepend-path      LD_LIBRARY_PATH   $root/lib
-prepend-path      CMAKE_MODULE_PATH   $root/share/ALPSCore


galexv:
According to [CMake docs for `find_package`](https://cmake.org/cmake/help/latest/command/find_package.html), the package file (`Find${name}.cmake`) is searched both in `${prefix}` and in `${prefix}/share/cmake/${name}` (among other places), and `$prefix` could be `${name}_DIR`, `CMAKE_PREFIX_PATH`, `PATH` and some other environment vars.

So, i am a bit surprised that setting environment variable `ALPSCore_DIR` to `$root` did not work for finding `$root/share/cmake/FindALPSCore.cmake`. For me (on a linux box), it works.

On the other hand, setting the *CMake* variable `ALPSCore_DIR` to `$root` won't work: it has to be set to the complete directory path.

ryanlevy:
Maybe this is a problem... I don't have that folder/file being generated. My share folder:
```
ALPSCore alps-accumulators alps-alea alps-gf  alps-hdf5 alps-mc alps-params alps-utilities
```
With install flags: 
``` 
-DCMAKE_INSTALL_PREFIX=/Users/[me]/modules/ALPSCore/clang/build -DCMAKE_BUILD_TYPE=Release -DEIGEN3_INCLUDE_DIR=$EIGEN3_INCLUDE_DIR -DDocumentation=OFF
```
The Jenkins build doesn't seem to have a FindALPSCore.cmake either.

galexv:
Eh, `${prefix}/share/${name}/` is also a search candidate. :) There is a file `${CMAKE_INSTALL_PREFIX}/share/ALPSCore/ALPSCoreConfig.cmake`.

But anyway, setting environment `ALPSCore_DIR` from the lmod script to the exact directory containing `ALPSCoreConfig.cmake` works too. (My point that a user may do only `export ALPSCore_DIR=/path/to/install_root` and it should work, saving some typing). Also, setting to the exact path allows a user to do something like `cmake -DALPSCore_DIR=${ALPSCore_DIR}`, which is redundant, but still should work.

What would be the purpose of `ALPSCore_ROOT`? Manual linking via `-L ${ALPSCore_ROOT}/lib` ?

ryanlevy:
If I set `export ALPSCore_DIR=$root`  then I get the error:
```
Could not find a package configuration file provided by "alps-params" with
  any of the following names:

    alps-paramsConfig.cmake
    alps-params-config.cmake
```
For example in the params tutorial, or in the linking tutorial:
```
 Could NOT find ALPSCore (missing: ALPSCore_LIBRARIES) (Required is at least
  version "2.0")
```
Should I open a bug report?

---

I like setting a root variable for access to the actual build/install directory if the `ALPSCore_DIR` setting is moved. Just for easy access potentially? I'm willing to remove it.

galexv:
> Should I open a bug report?

Yes, just to move the discussion to the convenient place for others to find.
Could you please also attach the exact sequence of commands and the `CMakeCache.txt` files, both for ALPSCore and for the tutorials?

ryanlevy:
Closed by #455

PR #441, opened 2018-01-31
[alea] Enables joining for different types

katherlee:
@mwallerb This is your fix for joining different types of results, and it has worked fine in my use case. Could you merge it once you think it's fine?


mwallerb:
@egull @katherlee This is fine, but still needs some work, I will do this as soon as the paper is done.

PR #442, opened 2018-01-31
Changed docs location to a more reasonable one.

yurivict:


galexv:
> Yes, this is a much better location w.r.t. the FHS.

Agreed. :)
It should me made configurable, anyway...

PR #446, opened 2018-01-31
ALEA: stream_serializer, a result serialization plugin

krivenko:
This plugin should allow for Boost/HPX serialization of `*_result<T>` types.
To enable the plugin for all supported ALEA result types in an HPX application one must include `alps/alea/plugin/stream_serializer.hpp` as follows,

```c++
#include <hpx/runtime/serialization/serialize.hpp>
#include <hpx/runtime/serialization/complex.hpp>

#include <alps/alea/plugin/stream_serializer.hpp>

namespace alps { namespace alea {
HPX_SERIALIZATION_SPLIT_FREE_TEMPLATE((template <typename T>), (mean_result<T>));
HPX_SERIALIZATION_SPLIT_FREE_TEMPLATE((template <typename T>), (var_result<T>));
HPX_SERIALIZATION_SPLIT_FREE_TEMPLATE((template <typename T>), (cov_result<T>));
HPX_SERIALIZATION_SPLIT_FREE_TEMPLATE((template <typename T>), (autocorr_result<T>));
HPX_SERIALIZATION_SPLIT_FREE_TEMPLATE((template <typename T>), (batch_result<T>));
}}
```

### DIFF
--- a/alea/include/alps/alea/autocorr.hpp
+++ b/alea/include/alps/alea/autocorr.hpp
@@ -199,6 +199,10 @@ public:
 
     level_result_type &level(size_t i) { return level_[i]; }
 
+    /** Check if this result is identical to another */
+    bool operator==(const autocorr_result &other) const;
+    bool operator!=(const autocorr_result &other) const { return !operator==(other); }
+
 protected:
     void reduce(const reducer &r, bool do_pre_commit, bool do_post_commit);
 
--- a/alea/include/alps/alea/autocorr.hpp
+++ b/alea/include/alps/alea/autocorr.hpp
@@ -209,6 +209,8 @@ private:
     friend class autocorr_acc<T>;
 };
 
+template<typename T> struct is_alea_result<autocorr_result<T>> : std::true_type {};
+
 template <typename T>
 struct traits< autocorr_result<T> >
 {
--- a/alea/include/alps/alea/autocorr.hpp
+++ b/alea/include/alps/alea/autocorr.hpp
@@ -199,10 +199,6 @@ public:
 
     level_result_type &level(size_t i) { return level_[i]; }
 
-    /** Check if this result is identical to another */
-    bool operator==(const autocorr_result &other) const;
-    bool operator!=(const autocorr_result &other) const { return !operator==(other); }
-
 protected:
     void reduce(const reducer &r, bool do_pre_commit, bool do_post_commit);
 
@@ -212,6 +208,15 @@ private:
 
     friend class autocorr_acc<T>;
 };
+
+/** Check if two results are identical */
+template <typename T>
+bool operator==(const autocorr_result<T> &r1, const autocorr_result<T> &r2);
+template <typename T>
+bool operator!=(const autocorr_result<T> &r1, const autocorr_result<T> &r2)
+{
+    return !operator==(r1, r2);
+}
 
 template<typename T> struct is_alea_result<autocorr_result<T>> : std::true_type {};
 
--- a/alea/include/alps/alea/plugin/stream_serializer.hpp
+++ b/alea/include/alps/alea/plugin/stream_serializer.hpp
@@ -85,7 +85,7 @@ private:
 };
 
 /**
- * stream_serializer
+ * stream_deserializer
  *
  * This class establishes connection between the ALEA serialization
  * interface and Boost/HPX serialization frameworks. Its instance
--- a/alea/include/alps/alea/plugin/stream_serializer.hpp
+++ b/alea/include/alps/alea/plugin/stream_serializer.hpp
@@ -22,13 +22,11 @@ namespace alps { namespace alea {
  * stream_serializer
  *
  * This class establishes connection between the ALEA serialization
- * interface and Boost/HPX serialization frameworks. Its instances
- * are used in free functions save() and load() (see below) that are
- * called upon Boost/HPX (de)serialization of ALEA *_result<T> types.
+ * interface and Boost/HPX serialization frameworks. Its instance
+ * is used in free function save() (see below) that is called upon
+ * Boost/HPX serialization of ALEA *_result<T> types.
  */
-template <typename Archive> class stream_serializer
-    : public serializer
-    , public deserializer
+template <typename Archive> class stream_serializer : public serializer
 {
 
 public:
@@ -64,7 +62,52 @@ public:
         do_write(data_view);
     }
 
+protected:
+
+    template <typename T> void do_write(const ndview<const T> &data_view)
+    {
+        const size_t * shape = data_view.shape();
+        size_t size = compute_size(shape, data_view.ndim());
+
+        const T * data = data_view.data();
+        for(long n = 0; n < size; ++n)
+            ar_ << *(data + n);
+    }
+
+    static size_t compute_size(const size_t *shape, size_t ndim)
+    {
+        return std::accumulate(shape, shape + ndim, 1, std::multiplies<size_t>());
+    }
+
+private:
+
+    Archive &ar_;
+};
+
+/**
+ * stream_serializer
+ *
+ * This class establishes connection between the ALEA serialization
+ * interface and Boost/HPX serialization frameworks. Its instance
+ * is used in free function load() (see below) that is called upon
+ * Boost/HPX deserialization of ALEA *_result<T> types.
+ */
+template <typename Archive> class stream_deserializer : public deserializer
+{
+
+public:
+
+    stream_deserializer(Archive &ar) : ar_(ar) {}
+
+    // Common methods
+
+    // Nothing to be done here: streams have no notion of groups
+    void enter(const std::string &) override {}
+    void exit() override {}
+
     // Deserialization methods
+
+    // Key names are irrelevant
 
     std::vector<size_t> get_shape(const std::string &key) override
     {
@@ -94,16 +137,6 @@ public:
 
 protected:
 
-    template <typename T> void do_write(const ndview<const T> &data_view)
-    {
-        const size_t * shape = data_view.shape();
-        size_t size = compute_size(shape, data_view.ndim());
-
-        const T * data = data_view.data();
-        for(long n = 0; n < size; ++n)
-            ar_ << *(data + n);
-    }
-
     template <typename T> void do_read(ndview<T> &data_view)
     {
         const size_t * shape = data_view.shape();
@@ -130,7 +163,6 @@ private:
     Archive &ar_;
 };
 
-
 /**
  * This method will be called by Boost/HPX Serialization libraries
  * when serialization of an ALEA result type is requested.
@@ -149,7 +181,7 @@ save(Archive &ar, const T &r, const unsigned int) {
 template <typename Archive, typename T>
 typename std::enable_if<is_alea_result<T>::value, void>::type
 load(Archive &ar, T &r, const unsigned int) {
-    stream_serializer<Archive> ser(ar);
+    stream_deserializer<Archive> ser(ar);
     deserialize(ser, "", r);
 }
 
--- a/alea/include/alps/alea/plugin/stream_serializer.hpp
+++ b/alea/include/alps/alea/plugin/stream_serializer.hpp
@@ -1,0 +1,165 @@
+/*
+ * Copyright (C) 1998-2018 ALPS Collaboration. See COPYRIGHT.TXT
+ * All rights reserved. Use is subject to license terms. See LICENSE.TXT
+ * For use in publications, see ACKNOWLEDGE.TXT
+ */
+#pragma once
+
+#include <numeric>
+#include <type_traits>
+
+#include <alps/alea/core.hpp>
+#include <alps/alea/complex_op.hpp>
+#include <alps/alea/mean.hpp>
+#include <alps/alea/variance.hpp>
+#include <alps/alea/covariance.hpp>
+#include <alps/alea/autocorr.hpp>
+#include <alps/alea/batch.hpp>
+
+namespace alps { namespace alea {
+
+/**
+ * stream_serializer
+ *
+ * This class establishes connection between the ALEA serialization
+ * interface and Boost/HPX serialization frameworks. Its instances
+ * are used in free functions save() and load() (see below) that are
+ * called upon Boost/HPX (de)serialization of ALEA *_result<T> types.
+ */
+template <typename Archive> class stream_serializer
+    : public serializer
+    , public deserializer
+{
+
+public:
+
+    stream_serializer(Archive &ar) : ar_(ar) {}
+
+    // Common methods
+
+    // Nothing to be done here: streams have no notion of groups
+    void enter(const std::string &) override {}
+    void exit() override {}
+
+    // Key names are irrelevant
+
+    void write(const std::string &, ndview<const double> data_view) override
+    {
+        do_write(data_view);
+    }
+    void write(const std::string &, ndview<const std::complex<double>> data_view) override
+    {
+        do_write(data_view);
+    }
+    void write(const std::string &, ndview<const complex_op<double>> data_view) override
+    {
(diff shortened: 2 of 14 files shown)


alea/include/alps/alea/plugin/stream_serializer.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include <alps/alea/mean.hpp>
+#include <alps/alea/variance.hpp>
+#include <alps/alea/covariance.hpp>
+#include <alps/alea/autocorr.hpp>
+#include <alps/alea/batch.hpp>
+
+namespace alps { namespace alea {
+
+// Metafunction to detect ALEA result types
+template<typename T> struct is_alea_result : std::false_type {};
```
krivenko: I thought we could put these 6 lines into main ALEA sources, a place like `alea/include/alps/alea/core.hpp`.
Or scatter them over `mean.hpp`, `variance.hpp` and so on...

mwallerb: Hmmm, right now we have this `traits<...>` struct; this is the C++98 "way" of doing these things ... I have the feeling It has since falled out of favor somewhat with C++11 and whatnot ... do you think we shall merge it into something like` `traits<...>::IS_RESULT` or do you think it is better separated out?

alea/include/alps/alea/plugin/stream_serializer.hpp
```diff
@@ -0,0 +1,84 @@
...
+    : public alps::alea::serializer
+{
+
+  stream_serializer(Archive &ar) : ar_(ar) {}
+
+  void write(const std::string &, const computed<double> &data) { do_write(data); }
+  void write(const std::string &, const computed<std::complex<double>> &data) { do_write(data); };
+  void write(const std::string &, const computed<complex_op<double>> &data) { do_write(data); };
+  void write(const std::string &, const computed<long> &data) { do_write(data); };
+  void write(const std::string &, const computed<unsigned long> &data) { do_write(data); };
```
mwallerb: These will have to change now, but no more allocation :-)
Also suggest `override`.

alea/include/alps/alea/plugin/stream_serializer.hpp
```diff
@@ -0,0 +1,84 @@
...
+void serialize(Archive &ar, complex_op<T> const &co, const unsigned int) {
+  ar & co.rere() & co.reim() & co.imre() & co.imim();
+}
+
+//
+// Serialize/deserialize *_result<T>
+//
+
+template <typename Archive> struct stream_serializer
+    : public alps::alea::serializer
```
mwallerb: Please add some documentation here.

alea/include/alps/alea/plugin/stream_serializer.hpp
```diff
@@ -0,0 +1,84 @@
...
+save(Archive &ar, T const &r, const unsigned int) {
+  stream_serializer<Archive> ser(ar);
+  serialize(ser, r);
+}
+
+template <typename Archive, typename T>
+typename std::enable_if<is_alea_result<T>::value, void>::type
+load(Archive &ar, T &r, const unsigned int) {
+  // TODO: deserialization
+}
```
mwallerb: I think documenting the rational these methods is important, particularly why they are needed next to the `serialize()` and `deserialize()` method

alea/test/stream_serializer.cpp
```diff
@@ -0,0 +1,77 @@
...
+        std::stringstream ss;
+        ss << in; // serialize
+
+        Acc out_acc(2);
+        auto out = out_acc.result();
+
+        ss >> out; // deserialize
+
+        // FIXME: How do I check that in == out in the generic way?
+        std::cout << ss.str() << std::endl;
```
mwallerb: We should provide an util for this.  One needs to be a little bit careful here though with a strict `==` comparison, because the stream you write to might be a text file and then you may very well lose a bit-by-bit equivalence after reading it back.  Probably something like Eigen's `isApprox()`, I'm thinking something along the lines:

    template <typename Res>
    alps::alea::util::is_approx(const Res &a, const Res &b, double atol);

alea/test/stream_serializer.cpp
```diff
@@ -0,0 +1,77 @@
...
+        //EXPECT_EQ(in.store().data(), out.store().data());
+    }
+};
```
mwallerb: Thanks for writing a test for this!

alea/include/alps/alea/plugin/stream_serializer.hpp
krivenko: > Hmmm, right now we have this traits<...> struct; this is the C++98 "way" of doing these things ... I have the feeling It has since falled out of favor somewhat with C++11 and whatnot ...

Trait bundles similar to `traits<...>` have become less popular for a reason. When you ask for `traits<mean_result<double>>::HAVE_MEAN` you make compiler spend time on instantiating all other members of the structure too.

> do you think we shall merge it into something liketraits<...>::IS_RESULT` or do you think it is better separated out?

Calling `traits<double>::IS_RESULT` would lead to a compilation error, unless you include `IS_RESULT` into the main declaration of `struct traits`. This is not very nice, I think...

alea/test/stream_serializer.cpp
krivenko: If we are to test serialization/deserialization, the comparison must be strict. Of course, different archive implementations may allow loss of precision. But in this particular test we need an archive implementation/emulation that is guaranteed to store the unaltered data (`std::stringstream` is probably not good enough for this...)

With this said, having `is_approx` would be nice indeed.

galexv:
> If we are to test serialization/deserialization, the comparison must be strict. [...] But in this particular test we need an archive implementation/emulation that is guaranteed to store the unaltered data (std::stringstream is probably not good enough for this...)

Would `std::stringstream` with [`std::hexfoat`](http://en.cppreference.com/w/cpp/io/manip/fixed) do? The hexfloat is an exact FP representation, if i am not mistaken.

krivenko:
@galexv Thank you Alex for the clue! I'll use `std::hexfoat` in the test...

mwallerb:
This is exactly my point. stringstream will not be compatible with ==

Markus

⁣Sent from BlueMail ​

krivenko:
@galexv Unfortunately, `std::hexfloat` is broken on gcc: https://stackoverflow.com/questions/42604596/read-and-write-using-stdhexfloat

galexv:
This is bad. :( Then probably a properly-behaving implementation of stream-like objects is indeed needed.  If all is needed to support `<<` and `>>`, it should not be very hard; for example, it can dump the exact memory content in an internal buffer, something along these lines:
```c++
/// Crude stream class implementation
struct dummy_stream {
    std::vector<char> buf_;

    void clear() { buf_.clear(); }

    template <typename T>
    dummy_stream& operator<<(const T& val) {
        std::copy((const char*)&val, (const char*)&val+sizeof(val), std::back_inserter(buf_));
        return *this;
    }

    template <typename T>
    dummy_stream& operator>>(T& val) {
        std::copy(buf_.data(), buf_.data()+sizeof(val), (char*)&val);
        return *this;
    }
};
```
(Of course, for real some type information, at least the number of bytes, must be saved too)
Does it make sense? Or am i missing the point completely?

krivenko:
>  If all is needed to support << and >>, it should not be very hard; for example, it can dump the exact memory content in an internal buffer, something along these lines:

Yes, I have already done something similar.
It's taking much more time to make that test work than to implement the serializer itself.

krivenko:
@mwallerb `stream_serializer` test fails for `autocorr_acc`. This happens because there is an `inf` and a `nan` in the input result object. Can you explain where they are coming from?

mwallerb:
Probably because the uppermost level has 1 measurement, which means that the variance is X/(1-1) = nan/inf (depending on whether X=0). Can you open an issue for this?

krivenko:
@mwallerb @galexv @egull @kleinhenz
This PR is ready for review, even though `stream_serializer` test is still affected by issue #462.

alea/test/stream_serializer.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+#include "gtest/gtest.h"
+#include "dataset.hpp"
+
+#include <vector>
+#include <queue>
+#include <cstdint>
+
+// Mock class that emulates Boost/HPX serialization archive
+class MockArchive {
```
galexv: I'd suggest to add some test(s) for `MockArchive` itself.

alea/test/stream_serializer.cpp
```diff
@@ -0,0 +1,141 @@
...
+    template<typename T>
+    void store_fundamental(T x) {
+        std::int8_t* p = reinterpret_cast<std::int8_t*>(&x);
+        for(int n = 0; n < sizeof(T); ++n)
+            buf.push(*(p + n));
+    }
+
+    template<typename T>
+    void extract_fundamental(T &x) {
+        std::int8_t* p = reinterpret_cast<std::int8_t*>(&x);
```
galexv: TL;DR;  At the risk of sounding pedantic, i suggest using `unsigned char` rather than `std::int8_t`

Rationale:
According to [type aliasing rules](http://en.cppreference.com/w/cpp/language/reinterpret_cast#Type_aliasing), accessing `x` via `p` is UB, unless `std::int8_t` happens to be a typedef for `unsigned char`, and it cannot be the case, because according to  http://en.cppreference.com/w/cpp/types/integer, `std::int8_t` is a signed type.

(As a side-note, I believe `std::uint8_t` is most probably a typedef for `unsigned char` on architectures we care about, although those "fixed width integer types" are technically optional.)

But we should use `unsigned char` regardless, because we really do not care if our `char` is 8 bits --- `sizeof` is defined in terms of `char`s, not 8-bit bytes.

kleinhenz:
On mac with clang 5.0.1 `stream_serializer` test immediately aborts with `libc++abi.dylib: terminating with uncaught foreign exception`.

stack trace:

```
frame #6: 0x00007fff6c9b7850 libc++abi.dylib`std::terminate() + 64
    frame #7: 0x000000010000531b stream_serializer`__clang_call_terminate + 11
    frame #8: 0x0000000100004a21 stream_serializer`twogauss_serialize_case_MockArchive_Test::TestBody() + 1025
    frame #9: 0x00000001006b7e98 libgtest.dylib`void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) + 72
    frame #10: 0x00000001006b7d6e libgtest.dylib`testing::Test::Run() + 334
    frame #11: 0x00000001006b8fe0 libgtest.dylib`testing::TestInfo::Run() + 304
    frame #12: 0x00000001006b99c7 libgtest.dylib`testing::TestCase::Run() + 311
    frame #13: 0x00000001006c2617 libgtest.dylib`testing::internal::UnitTestImpl::RunAllTests() + 1207
    frame #14: 0x00000001006c1fc8 libgtest.dylib`bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) + 72
    frame #15: 0x00000001006c1f37 libgtest.dylib`testing::UnitTest::Run() + 167
    frame #16: 0x000000010039df31 libgtest_main.dylib`main + 49
    frame #17: 0x00007fff6e914115 libdyld.dylib`start + 1
    frame #18: 0x00007fff6e914115 libdyld.dylib`start + 1
```

krivenko:
@kleinhenz Oh, that's nice. Now I have to fix a test for a class whose sole purpose is to test another class...

kleinhenz:
Sounds like it's time for `MockMockArchive`

krivenko:
@kleinhenz Any ideas? Must be something with the raw pointer arithmetics...

galexv:
@kleinhenz  @krivenko 
Could omitted `return *this` at https://github.com/ALPSCore/ALPSCore/blob/67bb06b2a87e59c8924090c60e5ccfeda51c8d73/alea/test/stream_serializer.cpp#L47-L52 be the culprit? If it ever invoked as `ar >> complex_number >> some_other_thing` ...

galexv:
@krivenko  You'll be laughing, but this seems to have been exactly the problem! In my standalone version of the test (with all references to ALEA and ALPSCore stripped off) clang-compiled code crashes with  `SIGILL` without the [commit e0daab3]  fix. Gcc-compiled code runs successfully.

kleinhenz:
That fixed it for me.

alea/test/stream_serializer.cpp
```diff
@@ -0,0 +1,167 @@
...
+        load(*this, result, 0);
+        return *this;
+    }
+
+private:
+
+    template<typename T>
+    void store_fundamental(T x) {
+        unsigned char* p = reinterpret_cast<unsigned char*>(&x);
+        for(int n = 0; n < sizeof(T); ++n)
```
galexv: It would also be good to use `std::size_t` rather than `int` here: one less warning with `-Wall`.

alea/test/stream_serializer.cpp
```diff
@@ -0,0 +1,167 @@
...
+            buf.push(*(p + n));
+    }
+
+    template<typename T>
+    void extract_fundamental(T &x) {
+        unsigned char* p = reinterpret_cast<unsigned char*>(&x);
+        for(int n = 0; n < sizeof(T); ++n) {
```
galexv: The same comment as above 
> It would also be good to use `std::size_t` rather than `int` here: one less warning with `-Wall`

alea/include/alps/alea/autocorr.hpp
```diff
@@ -199,6 +199,10 @@ class autocorr_result
 
     level_result_type &level(size_t i) { return level_[i]; }
 
+    /** Check if this result is identical to another */
+    bool operator==(const autocorr_result &other) const;
+    bool operator!=(const autocorr_result &other) const { return !operator==(other); }
```
mwallerb: A little bit of bikeshedding: I find non-inplace binary operators are conceptually clearer and more versatile as free functions.  Not a breaker though.
See also: http://www.drdobbs.com/cpp/how-non-member-functions-improve-encapsu/184401197

alea/test/stream_serializer.cpp
```diff
@@ -0,0 +1,167 @@
+#include <alps/alea/plugin/stream_serializer.hpp>
+
+#include "gtest/gtest.h"
+#include "dataset.hpp"
+
+#include <vector>
+#include <queue>
+
+// Mock class that emulates Boost/HPX serialization archive
+class MockArchive {
```
mwallerb: for consistency, this class should probably be called `mock_archive`

krivenko:
@mwallerb May I press the 'merge' button? I need these changes pretty much now...

mwallerb:
Course!

⁣Sent from BlueMail ​

alea/src/autocorr.cpp
```diff
@@ -104,6 +104,16 @@ void autocorr_acc<T>::finalize_to(autocorr_result<T> &result)
 template class autocorr_acc<double>;
 template class autocorr_acc<std::complex<double> >;
 
+template <typename T>
+bool autocorr_result<T>::operator==(const autocorr_result &other) const
+{
+    if(nlevel() != other.nlevel()) return false;
+    for(size_t i = 0; i < nlevel(); ++i) {
+        if(level(i) != other.level(i))
```
galexv: @krivenko  @mwallerb :  To address https://github.com/ALPSCore/ALPSCore/pull/446#issuecomment-365810182 and as a workaround for #462, could we compare `autocorr_result` equal even if contains `NaN` (or, at least, if it contains `NaN` at uppermost level)? This test failure is annoying. :)

krivenko: Let's just disable the test for the time being.

PR #447, opened 2018-02-02
return by const ref instead of value

kleinhenz:
`const` element access should probably return by `const ref` instead of by value. This is how it is done for both `std::vector` and `boost::multiarray`. I think it is more efficient as well as more flexible since it leaves it up to the consumer how they want to take the value.


krivenko:
Can we have this one merged or rejected?

I am trying to clean out this inchworm GF mess (which is already irritating as f*ck), and now this issue stands in my way.

Issue #453, opened 2018-02-04
[ALPSCore] [alps-gf-c++11] can't compile with gcc 4.7.2


iskakoff:
gcc-4.7.2 can't construct std::array using list initialization.

krivenko:
Strictly speaking, gcc 4.7 is not C++11-conformant.

https://gcc.gnu.org/gcc-4.8/cxx0x_status.html
```
As of GCC 4.8.1, GCC's C++11 mode implements all of the major features of the C++11 standard produced by the ISO C++ committee.
```
In practice, I would not recommend to use anything older than gcc 4.8.4, which had most of major c++11 bugs ironed out.

iskakoff:
@krivenko yes, but Emanuel has fedora 17 build server for some reason, and it has gcc-4.7.2. So far it was the only issue, I guess

iskakoff:
fixed by #454

Issue #450, opened 2018-02-02
[ALPSCore] [alea] [alps-gf-c++11] can't compile with intel 15


iskakoff:
Intel 15 on Mac is too stupid to properly understand c++11 templates features

iskakoff:
fixed by #454

PR #454, opened 2018-02-04
Compiler issues

iskakoff:
#450 and #453 issues


gf/include/alps/gf/gf_base.hpp
```diff
@@ -629,7 +629,9 @@ namespace alps {
          */
         template<size_t...Is>
         inline std::array < size_t, N_ > fill_sizes(const mesh_types &grids, index_sequence<Is...>) {
-          return {size_t(std::get<Is>(grids).extent())...};
```
krivenko: Will gcc 4.7 understand `return {{size_t(std::get<Is>(grids).extent())...}};`?

Issue #440, opened 2018-01-31
[alea] `autocorr_result` can fail to reduce when count is different between threads


katherlee:
This would cause the result to have different number of levels, and one can get a `MPI_ERR_TRUNCATE` or stuff.

Issue #431, opened 2018-01-23
[alea] Stream insertion (printing) for result types


krivenko:
Hi Markus,

Could you please consider adding `operator<<` to `alps::alea::var_result` and the likes?
It might be of use for debugging.

Thanks in advance!

Issue #61, opened 2014-10-26
Merging two accumulators


aeantipov:
When I do staging, I have old and new data - each being an alps::accumulator. What is the right way of merging the two?

galexv:
As of commit 3824d7b implemented for everything except FullBinningAccumulators.
I am moving this to a next release.

galexv:
Implementation for FullBinning needs proper testing, see issue #172.

egull:
@mwallerb can this be closed because Alea has corresponding functionality that works?

mwallerb:
No, not yet. "Merging" accumulators is IMHO a little bit of a misnomer, since it means adding the accumulated values of one accumulator to another.  After discussion with @kleinhenz and @krivenko, who need this, I'm working on implementing the following:

    mean_acc<T> &mean_acc<T>::operator<<(const mean_result<T> &);

PR #456, opened 2018-02-09
First overhaul of the ALEA library

mwallerb:
There are the following interface breaks:

  *  `sink<T>` has been renamed to `view<T>`
  *  `column<T>` is not directly convertible to `T` anymore.  

The scalars were removed, because they caused all sorts of problems when mixing them with Eigen functions etc.; down the line, we will introduce a wrapper for scalar accumulators to make working with them easier.

Fixes: #61
Fixes: #431
Fixes: #440


alea/include/alps/alea/batch.hpp
```diff
@@ -178,9 +190,15 @@ class batch_result
     /** Number of components of the random vector (e.g., size of mean) */
     size_t size() const { return store_->size(); }
 
+    /** Number of stores batches */
+    size_t num_batches() const { return store_->num_batches(); }
+
     /** Returns sample size, i.e., total number of accumulated data points */
     size_t count() const { return store_->count().sum(); }
 
+    /** Returns sample size, i.e., total number of accumulated data points */
```
egull: sample size or sample size ^2?

alea/include/alps/alea/covariance.hpp
```diff
@@ -229,10 +246,10 @@ class cov_result
     const column<T> &mean() const { return store_->data(); }
 
     /** Returns bias-corrected sample variance */
-    column<var_type> var() const { return store_->data2().diagonal().real(); }
+    column<var_type> var() const { return batch_size() * store_->data2().diagonal().real(); } // TODO
```
egull: Todo what?

alea/include/alps/alea/covariance.hpp
```diff
@@ -229,10 +246,10 @@ class cov_result
...
 
     /** Returns bias-corrected sample covariance matrix  */
-    const typename eigen<cov_type>::matrix &cov() const { return store_->data2(); }
+    typename eigen<cov_type>::matrix cov() const { return batch_size() * store_->data2(); } // TODO
```
egull: same as just above... Todo?

alea/include/alps/alea/hdf5.hpp
```diff
@@ -14,97 +15,210 @@
...
+            return std::vector<size_t>();
+        else
+            return archive_->extent(path);
     }
 
 private:
     hdf5::archive *archive_;
     std::string path_;
+    std::vector<std::string> group_;
+    bool debug_;
```
egull: do you like the debug and associated statements in the code?

egull:
...in addition.... tests?

egull:
Yep, that fixes the include issue. Have a look at this though:

[ 95%] Built target params_eq
Scanning dependencies of target twogauss
In file included from /Users/egull/ALPSCore/alea/test/mpi_twogauss.cpp:1:
/Users/egull/ALPSCore/alea/include/alps/alea/mpi.hpp:69:10: warning: 'commit' overrides a member function but is not marked 'override'
      [-Winconsistent-missing-override]
    void commit() const { }
         ^
/Users/egull/ALPSCore/alea/include/alps/alea/core.hpp:265:18: note: overridden virtual function is here
    virtual void commit() const = 0;
                 ^

alea/src/batch.cpp
```diff
@@ -89,10 +92,27 @@ void batch_acc<T>::add(const computed<T> &source, size_t count)
...
+batch_acc<T> &batch_acc<T>::operator<<(const batch_result<T> &other)
+{
+    internal::check_valid(*this);
+    if (size() != other.size())
+        throw size_mismatch();
+    // TODO this is not strictly speaking necessary when done properly
+    if (num_batches() != other.num_batches())
+        throw size_mismatch();
+
+    // FIXME: this is a terrible idea because it mixes two time series
```
krivenko: Just forbid merging results into `batch_acc` then?

alea/src/batch.cpp
```diff
@@ -216,29 +236,78 @@ template column<double> batch_result<double>::var<circular_var>() const;
...
+    deserialize(s, "@num_batches", new_nbatches);
+    if (!self.valid() || self.size() != new_size || self.store().num_batches() != new_nbatches)
+        self.store_.reset(new batch_data<T>(new_size, new_nbatches));
+
+    // deserialize data
+    s.enter("batch");
+    deserialize(s, "count", self.store().count());
+    deserialize(s, "sum", self.store().batch());
+    s.exit();
+}
```
krivenko: What about deserializing `mean`?

alea/src/covariance.cpp
```diff
@@ -79,13 +82,32 @@ template <typename T, typename Str>
...
 
+template <typename T, typename Str>
+cov_acc<T,Str> &cov_acc<T,Str>::operator<<(const cov_result<T,Str> &other)
+{
+    internal::check_valid(*this);
+    if (size() != other.size())
+        throw size_mismatch();
+
+    // NOTE partial sums are unchanged
+    // HACK we need this for "outwardly constant" manipulation
```
krivenko: This is ugly indeed... Can you think of a better approach?

alea/src/covariance.cpp
```diff
@@ -190,23 +212,66 @@ template class cov_result<std::complex<double>, elliptic_var>;
...
-    s.write("mean/value", make_adapter(self.mean()));
-    s.write("mean/error", make_adapter(self.stderror()));
+    internal::serializer_sentry group(s, key);
+
+    serialize(s, "@size", self.store_->data_.size());
+    serialize(s, "count", self.store_->count_);
+    serialize(s, "count2", self.store_->count2_);
+    s.enter("mean");
+    serialize(s, "value", self.store_->data_);
+    serialize(s, "error", self.stderror());   // TODO temporary
```
krivenko: I understand that in general numbers of calls to `serialize()` and `deserialize()` don't have to match.
However, how am I going to implement `stream_(de)serializer` (#446)?
All it would be able to do is to write/read the next value to/from stream ignoring key names altogether.

alea/include/alps/alea/core.hpp
```diff
@@ -37,39 +37,95 @@ template <typename T>
...
+/**
+ * Data view as a thin wrapper around a continuous multi-dimensional array.
+ *
+ * Basically collects a pointer continuous array in ROW-MAJOR format (vector,
+ * transposed Eigen array, C array etc.) together with its shape.  Note that
+ * the view neither owns the `data` pointer nor the `shape` pointer.  Thus it
+ * amends `view<T>` with shape information.
+ */
+template <typename T>
+class ndview
```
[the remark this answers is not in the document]
mwallerb: I disagree.  The dimension is a compile-time constant whenever it is or supposed to be a constant at compile time (this is important for multiarrays and Eigen tensors, because loop unrolling is crucial in dereferencing).  Here, the situation is different, since shape is essentially metadata to be attached to datasets.

mwallerb:
@egull 

> Scanning dependencies of target twogauss
> In file included from /Users/egull/ALPSCore/alea/test/mpi_twogauss.cpp:1:
> /Users/egull/ALPSCore/alea/include/alps/alea/mpi.hpp:69:10: warning: 'commit' overrides a member function but is not marked 'override'
> [-Winconsistent-missing-override]
> void commit() const { }
> ^

Should be fixed.

> In addition ... tests?

I have amended some of the tests to test the new functionality.  Also, I have added a test for the autocorr bug.   There is of course no comprehensive suite yet, but that is hopefully to come.

alea/include/alps/alea/internal/format.hpp
```diff
@@ -0,0 +1,190 @@
...
+    static T *& get_format(std::ios_base &stream)
+    {
+        return reinterpret_cast<T *&>(stream.pword(get_xindex()));
+    }
+
+    static int get_xindex()
+    {
+        // guaranteed to be distinct since for different iomanip's since their
+        // T will be different.
+        static int xindex = std::ios_base::xalloc();
```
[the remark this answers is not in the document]
mwallerb: > This call is a bit unsettling...
> http://en.cppreference.com/w/cpp/io/ios_base/xalloc
> This function is thread-safe; concurrent access by multiple threads does not result in a data race

I don't know if it is that much of a problem in practice -- any implementation I've seen of xalloc is thread-safe because this thing has been around since C++98 and otherwise all uses of the STL in the multi-threaded context would immediately blow up.

alea/include/alps/alea/serialize.hpp
```diff
@@ -0,0 +1,178 @@
...
+{
+    std::array<size_t, 2> shape = {{(size_t)value.cols(), (size_t)value.rows()}};
+    ser.read(key, ndview<T>(value.data(), shape.data(), shape.size()));
+}
+
+template <typename T>
+void deserialize(deserializer &ser, const std::string &key,
+                 Eigen::Matrix<T, Eigen::Dynamic, 1> &value)
+{
+    std::array<size_t, 1> shape = {{(size_t)value.rows()}};
```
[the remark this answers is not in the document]
mwallerb: >Serialization procedure seems to erase the difference between Eigen column-vectors and row-vectors?
Is that right?

Yes, that is deliberate.  In a serialized format, we usually don't care about this, and at deserialization it is dictated by the data structure we put it in anyway.

alea/src/batch.cpp
mwallerb: > Just forbid merging results into batch_acc then?

Hmmm, I would like to keep it because the mean is still computed correctly, and in principle it is sound to merge to batch accumulators.   The variance will be strange if you have correlated batches then, but the same issue is present with `reduce()` right now.

alea/src/covariance.cpp
mwallerb: > This is ugly indeed... Can you think of a better approach?

Yes, but I would like to wait until I have refactored these classes into a base class and a set of accumulator classes for different dimensionality.  Otherwise I have to do the work twice. I don't think it is a big issue though.

alea/src/covariance.cpp
mwallerb: > I understand that in general numbers of calls to serialize() and deserialize() don't have to match.
However, how am I going to implement stream_(de)serializer (#446)?
All it would be able to do is to write/read the next value to/from stream ignoring key names altogether.

Very good point. Please look at my fix for this.

alea/include/alps/alea/internal/format.hpp
```diff
@@ -8,6 +8,7 @@
 #include <cassert>
 #include <ostream>
 #include <memory>
+#include <mutex>
```
krivenko: I guess you wanted to protect that call to `std::ios_base::xalloc()` but then changed your mind...

mwallerb: Guilty as charged 😏 . I got confused about `mutex`es and then thought I better leave it before I make matters worse -- let's maybe discuss tomorrow.

alea/include/alps/alea/core.hpp
```diff
@@ -37,39 +37,95 @@ template <typename T>
...
+        , ndim_(ndim)
+    { }
+
+    /** Construct view on data area with shape and size hint */
+    ndview(T *data, size_t size, const size_t *shape, size_t ndim)
+        : view<T>(data, size)
+        , shape_(shape)
+        , ndim_(ndim)
+    {
+        assert(size == compute_size(shape, ndim));
```
iskakoff: Why do you need this constructor?

alea/include/alps/alea/core.hpp
```diff
@@ -37,39 +37,95 @@ template <typename T>
...
+{
+public:
+    typedef T value_type;
+
+public:
+    /** Construct view on nothing */
+    ndview() : view<T>(), shape_(nullptr), ndim_(0) { }
+
+    /** Construct view on data area with shape */
+    ndview(T *data, const size_t *shape, size_t ndim)
```
iskakoff: const size_t * shape? I don't like pointer here.
why not array or vector?:

    template<size_t N>
    void compute(T *data, const std::array<size_t, N> & shape, size_t ndim) {

alea/include/alps/alea/core.hpp
```diff
@@ -37,39 +37,95 @@ template <typename T>
...
+    ndview(T *data, size_t size, const size_t *shape, size_t ndim)
+        : view<T>(data, size)
+        , shape_(shape)
+        , ndim_(ndim)
+    {
+        assert(size == compute_size(shape, ndim));
+    }
+
+    /** Get shape of data space */
+    const size_t *shape() const { return shape_; }
```
iskakoff: and again, why not array or vector? it at least will guarantee, that in `compute_size` you will not go out of shape allocation

alea/include/alps/alea/core.hpp
```diff
@@ -217,49 +252,121 @@ struct reducer
...
+    virtual reducer *clone() { throw unsupported_operation(); }
+
     /** Destructor */
     virtual ~reducer() { }
 
     // Convenience functions
 
-    void reduce(sink<std::complex<double> > data) const {
-        reduce(sink<double>((double *)data.data(), 2 * data.size()));
+    void reduce(view<std::complex<double> > data) const {
```
iskakoff: what? why do you cast complex to pair of double? At least in case of mpi reduce you definitely shouldn't do this.

and why it's not reference?

alea/src/variance.cpp
```diff
@@ -171,10 +192,10 @@ void var_result<T,Str>::reduce(const reducer &r, bool pre_commit, bool post_comm
     internal::check_valid(*this);
     if (pre_commit) {
         store_->convert_to_sum();
-        r.reduce(sink<T>(store_->data().data(), store_->data().rows()));
-        r.reduce(sink<var_type>(store_->data2().data(), store_->data2().rows()));
-        r.reduce(sink<double>(&store_->count(), 1));
-        r.reduce(sink<double>(&store_->count2(), 1));
+        r.reduce(view<T>(store_->data().data(), store_->data().rows()));
+        r.reduce(view<var_type>(store_->data2().data(), store_->data2().rows()));
+        r.reduce(view<double>(&store_->count(), 1));
```
iskakoff: count is double? :hurtrealbad:

alea/include/alps/alea/autocorr.hpp
```diff
@@ -20,7 +20,13 @@ namespace alps { namespace alea {
     template <typename T> class batch_result;
 
     template <typename T>
-    void serialize(serializer &, const autocorr_result<T> &);
+    void serialize(serializer &, const std::string &, const autocorr_result<T> &);
```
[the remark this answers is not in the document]
mwallerb: This is just the forward declaration. The actual declarations come with documentation and parameter names

alea/include/alps/alea/core.hpp
mwallerb: view is a wrapper pointer, so you don't need to pass it by reference.

alea/include/alps/alea/core.hpp
mwallerb: Well, essentially because of performance consideration (I wanted to avoid unnecessary malloc's and the number of dimensions is not fixed at compile time). The main point is that there is an intrinsic scoping issue here anyway, because ndview, similar to, say, Eigen::Map, does not own its data pointer. Thus, while I can make the shape a std::vector, it does not make the scoping issue go away.

alea/src/variance.cpp
mwallerb: Well, yes, for consistency with count2_, as the sum of squares can easily overflow a long.

alea/include/alps/alea/core.hpp
mwallerb: I thought I add it for convenience.  I can remove it if it confuses things.

alea/include/alps/alea/core.hpp
mwallerb: And why cannot I cast a complex to a pair of double? A complex literally *is* a pair of doubles, is it not?  And reduction of a complex is the reduction of real and imaginary part, is it not?

alea/include/alps/alea/autocorr.hpp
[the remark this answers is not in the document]
iskakoff: where? autocorr.hpp has this forward declaration and then friend declaration, none of them come with parameter names. the same for other serializable classes

alea/include/alps/alea/core.hpp
iskakoff: Sure you can, but why? The only reason if your underlying library for reduction does not support complex, but at least mpi is not the case.

alea/src/variance.cpp
iskakoff: because in other places count is size_t

alea/include/alps/alea/core.hpp
iskakoff: the data pointer is 100% ok, but shape?
It will make life at least easy, because, finally many people will declare shape as something like a vector and then pass as .data(), .size(). 

another question, am I wrong, that you can't return view from the function if you construct shape inside this function? something like this:

     ndview create(blah blah blah) {
         std::vector<size_t> shape{{1,2,3,4}};
         return ndview(data, shape.data(), shape.size());
     }

PR #459, opened 2018-02-12
resizable green's functions

iskakoff:
make GF to be resizeable


gf/test/gf_new_test.cpp
```diff
@@ -317,3 +317,20 @@ TEST(GreensFunction, ConstRef) {
...
+TEST(GreensFunction, Reshape) {
+  alps::gf::matsubara_positive_mesh x(100, 10);
+  alps::gf::index_mesh y(10);
+  alps::gf::index_mesh y2(1);
+  alps::gf::index_mesh z(10);
+  alps::gf::index_mesh z2(100);
+  greenf<double, alps::gf::matsubara_positive_mesh, alps::gf::index_mesh, alps::gf::index_mesh > g1(x, y, z);
+  greenf<double, alps::gf::matsubara_positive_mesh, alps::gf::index_mesh, alps::gf::index_mesh > g2;
+  ASSERT_TRUE(g2.data().size() == 0);
+  std::array<size_t, 3> shape{{0}};
```
galexv: Minor point, but why not `{{0,0,0}}` (for clarity) or `{{ }}` (to default-initialize everything) ?

gf/include/alps/gf/gf_base.hpp
```diff
@@ -448,6 +448,26 @@ namespace alps {
                                             [](VTYPE a, VTYPE b) {return std::abs(a) < std::abs(b);} ) );
         }
 
+        // reshape green's function
+        template<typename St = Storage>
```
krivenko: I don't like those artificial template parameters needed only to make `std::enable_if` work.
In this simple case you could go with something cleaner:
```c++
        // reshape green's function
        void reshape(MESHES...meshes) {
          std::array<size_t, N_> new_shape = get_sizes(std::make_tuple(meshes...));

          if(std::is_same<data_view, Storage>::value) {
            size_t new_size = std::accumulate(new_shape.begin(), new_shape.end(), size_t(1), std::multiplies<size_t>());
            if(data_.size() != new_size) {
              throw std::invalid_argument("The total size of resulting Green's function view should be the same as before.");
            }
          }

          data_.reshape(new_shape);
          meshes_ = std::make_tuple(meshes...);
          empty_ = false;
        }
```

utilities/include/alps/numeric/tensors/tensor_base.hpp
```diff
@@ -406,6 +406,14 @@ namespace alps {
         /// sizes for each dimension
         const std::array < size_t, Dim > &shape() const { return shape_; };
 
+        /// reshape with index list
+        template<typename Ind, typename ...Inds>
+        typename std::enable_if<std::numeric_limits<Ind>::is_integer>::type reshape(Ind ind, Inds...inds) {
```
krivenko: You can check all index types using the following metafunction:
```c++
#include <iostream>
#include <type_traits>

template<typename I0, typename... I> struct are_all_integers :
    std::integral_constant<bool, std::is_integral<I0>::value && are_all_integers<I...>::value>
{};
template<typename I0> struct are_all_integers<I0> : std::is_integral<I0> {};

int main()
{   
    std::cout << std::boolalpha << are_all_integers<int>::value << std::endl;
    std::cout << std::boolalpha << are_all_integers<int,char,unsigned long>::value << std::endl;
    std::cout << std::boolalpha << are_all_integers<int,double,unsigned long>::value << std::endl;
    
    return 0;
}
```
```
true
true
false
```

kleinhenz: There is also [`std::conjunction`](http://en.cppreference.com/w/cpp/types/conjunction) which we could provided in utilities.

krivenko:
Could you please open another PR for #471 fix?

iskakoff:
@krivenko I wanted to commit them together

PR #463, opened 2018-02-17
RFC: [alea] add support for accumulating std::array

kleinhenz:
This adds support for accumulating `std::array`.

It also introduces `is_alea_acc` metafunction and makes insertion operators for non `computed` types into free functions defined in `computed.hpp` where the adapters are defined. I think this makes sense because it removes the 5x duplication of these functions in the definition of every accumulator, shortens the actual accumulator interfaces and doesn't special case any types. The user can add support for accumulating custom types by just following how it is done for `std` types in `computed.hpp`. The only downside I can think of is that now these functions are not explicitly instantiated with the accumulators. However since they are basically one-liners this is probably not too big of a deal.

This is done on top of #446. I will retarget to master when that is merged.

### DIFF
--- a/alea/include/alps/alea/autocorr.hpp
+++ b/alea/include/alps/alea/autocorr.hpp
@@ -67,8 +67,8 @@ class autocorr_acc
 {
 public:
     using value_type = T;
-    typedef typename bind<circular_var, T>::var_type var_type;
-    typedef var_acc<T, circular_var> level_acc_type;
+    using var_type = typename bind<circular_var, T>::var_type;
+    using level_acc_type = var_acc<T, circular_var>;
 
 public:
     autocorr_acc(size_t size=1, size_t batch_size=1, size_t granularity=2);
--- a/alea/include/alps/alea/autocorr.hpp
+++ b/alea/include/alps/alea/autocorr.hpp
@@ -66,7 +66,7 @@ template <typename T>
 class autocorr_acc
 {
 public:
-    typedef T value_type;
+    using value_type = T;
     typedef typename bind<circular_var, T>::var_type var_type;
     typedef var_acc<T, circular_var> level_acc_type;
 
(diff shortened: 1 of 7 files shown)


alea/include/alps/alea/autocorr.hpp
```diff
@@ -66,7 +66,7 @@ template <typename T>
 class autocorr_acc
 {
 public:
-    typedef T value_type;
+    using value_type = T;
     typedef typename bind<circular_var, T>::var_type var_type;
     typedef var_acc<T, circular_var> level_acc_type;
```
mwallerb: Why this and not the other typedef's here?

alea/include/alps/alea/variance.hpp
```diff
@@ -171,7 +160,7 @@ class var_acc
     /** Return backend object used for storing estimands */
     const var_data<T,Strategy> &store() const { return *store_; }
 
-protected:
+  protected:
```
mwallerb: Please remove the indentation for consistency :)

alea/include/alps/alea/autocorr.hpp
kleinhenz: No real reason. `mean_acc` was missing the `typedef` so I put a `using` expression and then wanted to make it consistent across all the accumulators. I can update rest of the `typedefs` if you want or switch `value_type` back.  I overall prefer the `using` syntax but it doesn't matter too much.

mwallerb: Yeah, let's do that then.

kleinhenz: switch to `using` or back to `typedef`?

mwallerb: switch to `using` -- I feel like this is the way the kids call it these days :-)

kleinhenz:
This still works

```
alps::alea::autocorr_acc<double> acc (4);
Eigen::VectorXd v (4);
acc << (v + 2 * v);
```

is this what you were concerned with?

krivenko:
I'd very much like this merged.
Right now I have to write
```c++
Eigen::Matrix<std::complex<double>, Eigen::Dynamic, 1> c(2);
c << sign, sign_norm;
acc << c;
```
while I could be writing
```c++
acc << std::array<std::complex<double>, 2>{sign, sign_norm};
```

mwallerb:
Ok, let's ship it then! (You don't have to wait for me alwa

PR #468, opened 2018-02-21
[alea] add std::initializer_list constructor to column

kleinhenz:
adds `std::initializer_list` constructor to column

### DIFF
--- a/alea/include/alps/alea/core.hpp
+++ b/alea/include/alps/alea/core.hpp
@@ -11,6 +11,7 @@
 #include <complex>
 #include <string>
 #include <vector>
+#include <initializer_list>
 
 #include <Eigen/Dense>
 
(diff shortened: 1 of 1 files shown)


krivenko:
Here is my use case:
```c++
struct transformer_ratio : public transformer<std::complex<double>>
{
    /** apply transformation */
    column<std::complex<double>> operator() (const column<std::complex<double>> &in) const {
      return column<std::complex<double>>{in(0) / in(1)};
    }
...
```

alea/include/alps/alea/core.hpp
```diff
@@ -188,6 +189,14 @@ class column
 
     column(size_t size) : Eigen::Matrix<T, Eigen::Dynamic, 1>(size) {}
 
+    column(std::initializer_list<T> l) : column(l.size()) {
+      int i = 0;
```
krivenko: `size_t`, please, to keep it consistent with the rest of the library

Issue #471, opened 2018-02-22
[ALEA] Compilation failure with OpenMPI 1.6


krivenko:
OpenMPI 1.6 declares the first argument of `MPI_Reduce` as `void*` while it should normally be `const void*` (contents of the send buffer is not supposed to be changed).
```
In file included from /home/travis/build/krivenko/ALPSCore/alea/test/mpi_twogauss.cpp:1:
/home/travis/build/krivenko/ALPSCore/alea/include/alps/alea/mpi.hpp:90:22: error: 
      no matching function for call to 'MPI_Reduce'
        mpi::checked(MPI_Reduce(sendbuf, data.data(), data.size(), dtype_tag,
                     ^~~~~~~~~~
/home/travis/build/krivenko/ALPSCore/alea/include/alps/alea/mpi.hpp:65:53: note: 
      in instantiation of function template specialization
      'alps::alea::mpi_reducer::inplace_reduce<double>' requested here
    void reduce(view<double> data) const override { inplace_reduce(data); }
                                                    ^
/usr/lib/openmpi/include/mpi.h:1352:20: note: candidate function not viable: 1st
      argument ('const void *') would lose const qualifier
OMPI_DECLSPEC  int MPI_Reduce(void *sendbuf, void *recvbuf, int count,
                   ^
In file included from /home/travis/build/krivenko/ALPSCore/alea/test/mpi_twogauss.cpp:1:
/home/travis/build/krivenko/ALPSCore/alea/include/alps/alea/mpi.hpp:90:22: error: 
      no matching function for call to 'MPI_Reduce'
        mpi::checked(MPI_Reduce(sendbuf, data.data(), data.size(), dtype_tag,
                     ^~~~~~~~~~
/home/travis/build/krivenko/ALPSCore/alea/include/alps/alea/mpi.hpp:67:51: note: 
      in instantiation of function template specialization
      'alps::alea::mpi_reducer::inplace_reduce<long>' requested here
    void reduce(view<long> data) const override { inplace_reduce(data); }
                                                  ^
/usr/lib/openmpi/include/mpi.h:1352:20: note: candidate function not viable: 1st
      argument ('const void *') would lose const qualifier
OMPI_DECLSPEC  int MPI_Reduce(void *sendbuf, void *recvbuf, int count,
                   ^
2 errors generated.
```

mwallerb:
Yeah, that sucks. (Also: 1.6?! Hello, Marty McFly!)

[...]

PR #474, opened 2018-02-23
issue #471

iskakoff:


alea/include/alps/alea/mpi.hpp
```diff
@@ -87,7 +87,7 @@ struct mpi_reducer
 
         // In-place requires special value for sendbuf, but only on root
         const void *sendbuf = am_root() ? MPI_IN_PLACE : data.data();
-        mpi::checked(MPI_Reduce(sendbuf, data.data(), data.size(), dtype_tag,
+        mpi::checked(MPI_Reduce(const_cast<void*>(sendbuf), data.data(), data.size(), dtype_tag,
```
galexv: IMHO, it would be good to leave a comment that `const_cast` is needed here because of problems with MPI-2.1, but the cast is in fact safe. 

(Btw, if you'll make a commit adding the comment, please [add `[ci skip]`](https://docs.travis-ci.com/user/customizing-the-build/#Skipping-a-build) to the commit message so that Travis won't try to test it again!)

Issue #462, opened 2018-02-15
[ALEA] Inf/NaN in autocorr_acc result


krivenko:
Division by zero occurs when `autocorr_acc` tries to normalize variance at the uppermost level.

https://github.com/ALPSCore/ALPSCore/pull/446#issuecomment-365810182

```c++
#include <iostream>
#include <vector>

#include <alps/alea/autocorr.hpp>
#include "dataset.hpp"

int main()
{
    alps::alea::autocorr_acc<double> acc(2);
    for (size_t i = 0; i != twogauss_count; ++i)
        acc << std::vector<double>{twogauss_data[i][0], twogauss_data[i][1]};

    alps::alea::autocorr_result<double> res = acc.result();

    std::cout << "nlevel = " << res.nlevel() << std::endl;
    std::cout << "level(9) = " << res.level(9) << std::endl;

    return 0;
}
```
```
nlevel = 10
level(9) =   1.00718
-0.497835 +-  inf
-nan
```

mwallerb:
Thinking some more about this, I think the bug is actually in `var_result::operator==`: two results with 1 element each should both have infinite variance and thus compare as equal even though `!(inf == inf)`

krivenko:
Okay, maybe it's a good idea to specifically test for variance being infinite, even though `inf == inf` on most platforms.
The real problem is caused by the NaNs, because `!(nan == nan)`. Those are not supposed to ever appear in any results, are they?

PR #475, opened 2018-02-23
[ALEA] Disable stream_serializer test until #462 is fixed

krivenko:
Just don't forget to switch it back...


galexv:
Thanks!
I'd rather put `DISABLED_` in front of test name, and also do some cosmetic changes. I'll create a PR.

PR #477, opened 2018-02-25
[alea] Handle the case of zero/one measurements reponsibly.

### LINKED ISSUE #462: [ALEA] Inf/NaN in autocorr_acc result

mwallerb:
Fixes #462

### DIFF
--- a/alea/include/alps/alea/complex_op.hpp
+++ b/alea/include/alps/alea/complex_op.hpp
@@ -201,6 +201,29 @@ public:
                           inv_t * x.imre(), inv_t * (x.imim() + s));
     }
 
+    friend bool isnan(complex_op x)
+    {
+        return std::any_of(&x.vals_[0][0], &x.vals_[2][0],
+                           static_cast<bool(*)(T)>(std::isnan));
+    }
+
+    friend bool isfinite(complex_op x)
+    {
+        return std::all_of(&x.vals_[0][0], &x.vals_[2][0],
+                           static_cast<bool(*)(T)>(std::isfinite));
+    }
+
+    friend bool isinf(complex_op x)
+    {
+        // An object cannot be inf and nan at the same time, for multi-
+        // component objects nan takes precedence
+        if (isnan(x))
+            return false;
+
+        return std::any_of(&x.vals_[0][0], &x.vals_[2][0],
+                           static_cast<bool(*)(T)>(std::isinf));
+    }
+
     friend complex_op abs(complex_op x) { return sqrt(abs2(x)); }
 
     friend std::ostream &operator<<(std::ostream &out, complex_op x)
--- a/alea/src/variance.cpp
+++ b/alea/src/variance.cpp
@@ -203,6 +203,9 @@ var_result<T,Str> &var_result<T,Str>::operator=(const var_result &other)
 template <typename T, typename Strategy>
 bool operator==(const var_result<T, Strategy> &r1, const var_result<T, Strategy> &r2)
 {
+    if (r1.count() == 0 && r2.count() == 0)
+        return true;
+
     return r1.count() == r2.count()
         && r1.count2() == r2.count2()
         && r1.store().data() == r2.store().data()
--- a/alea/src/variance.cpp
+++ b/alea/src/variance.cpp
@@ -5,6 +5,8 @@
 #include <alps/alea/internal/util.hpp>
 #include <alps/alea/internal/format.hpp>
 
+#include <limits>
+
 namespace alps { namespace alea {
 
 template <typename T, typename Str>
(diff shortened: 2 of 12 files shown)


alea/include/alps/alea/complex_op.hpp
```diff
@@ -201,6 +201,29 @@ class complex_op
                           inv_t * x.imre(), inv_t * (x.imim() + s));
     }
 
+    friend bool isnan(complex_op x)
```
egull: please comment your functions

alea/include/alps/alea/complex_op.hpp
```diff
@@ -201,6 +201,29 @@ class complex_op
...
+    }
+
+    friend bool isinf(complex_op x)
+    {
+        // An object cannot be inf and nan at the same time, for multi-
+        // component objects nan takes precedence
+        if (isnan(x))
+            return false;
+
+        return std::any_of(&x.vals_[0][0], &x.vals_[2][0],
```
egull: are you sure about the 2? Should it be a 1? (same everywhere else in this block)

mwallerb: No it's 2 here (ranges are exclusive).

alea/src/variance.cpp
```diff
@@ -27,17 +27,38 @@ void var_data<T,Str>::reset()
 template <typename T, typename Str>
 void var_data<T,Str>::convert_to_mean()
 {
+    // This also works for count_ == 0
```
krivenko: Hmm. What's gonna happen if `count_ == 0`?
`data_`, `data2_` and `nunbiased` will all be nan, no?

mwallerb: Exactly: `data_` and `data2_`, that is, the mean and variance will be `nan`.  This is correct conceptually, since both are `0./0.`.

krivenko: Fine. I agree.
Could you please amend `operator==` to return `true` immediately if `(r1.count() == 0) && (r2.count() == 0)`?

mwallerb: Oh yes, good point.

galexv: Along the same lines, and also as Emanuel(@egull) mentioned: there are no tests for `operator==`: how do we know that it works? ;)  And we have other tests that depend on it working.
(But possibly it should be a separate issue deserving a separate PR)

krivenko:
@mwallerb Just in case, it seems like we are not directly affected by this issue at the moment. No rush, therefore.

krivenko:
@mwallerb I tried to rebuild InchwormHPX on top of this PR.
If I am not mistaken, the estimated autocorrelation time is 0 after 1000 measurements because `DEFAULT_MIN_SAMPLES==1024`. Convergence of `tau` is indeed observed, but for much larger numbers of measurements.
```
n measurements  tau
--------------  ---
1000000         25.83
2000000         29.08
3000000         30.34
5000000         33.12
10000000        33.78
```

mwallerb:
Added some tests (and a little cleanup).

alea/test/stream_serializer.cpp
```diff
@@ -142,6 +144,7 @@ class twogauss_serialize_case
 
         archive >> out; // deserialize
 
+        std::cerr << alps::alea::PRINT_VERBOSE << "\nout\n" << out << "\n";
         EXPECT_EQ(in, out);
```
galexv: (Quoting https://github.com/ALPSCore/ALPSCore/pull/477#commitcomment-27816874):
> K: Just in case, why didn't you do the same for `batch_result`?
> M: Because `batch_result` does not normalize any results, i.e., the comparison works for `count() == 0`

Sorry for harping on this again :), but how do we know it?
It would be nice to have a test verifying that `in==out` is true iff `in` and `out` are equal, for usual cases and some reasonable corner cases like `count()==0` .

mwallerb: Let's make this a seperate issue though, shall we (#482)? Otherwise this PR will drag on forever ...

Issue #172, opened 2015-11-03
Set up proper statistical tests for accumulators


galexv:
What are the "correct" values for the error bars and autocorrelation lengths for accumulators? How to test them properly? 
Ensure that the "correct" values are printed out too (Cf issue #83, issue #129).

egull:
Markus, I'm assigning this to you. It's low priority but would be a good use of hypothesis testing.

mwallerb:
Fine.

Proposed Roadmap: 
* Set up a linear autoregressive process (AR(1))
* Bin it
* In this case the integrated and exponential autocorrelation time coincide with the characteristic relaxation of the process -- verify it stochastically.

galexv:
There is a code in [`accumulators/test/accumulator_generator.hpp`](https://github.com/ALPSCore/ALPSCore/blob/master/accumulators/test/accumulator_generator.hpp)
 (admittedly, wanting some refactoring) that [generates autocorrelated data](https://github.com/ALPSCore/ALPSCore/blob/master/accumulators/test/accumulator_generator.hpp#L124): in my understanding, it implements [AR(L)](https://en.wikipedia.org/wiki/Autoregressive_model#Definition) with equal weights (φᵢ=`1/(L+1)` ) or something reasonably close to it. May it be used as a start?

PR #488, opened 2018-03-14
[alea] add Hypothesis testing support

mwallerb:
Adds `alps::alea::test_mean(x, y)`, which takes either one statistical result and a benchmark or two statistical results, and returns the corresponding p-value.

For testing adds
  * VAR(1) model (`alps::alea::util::var1_model`), a simple Markov process with a finite autocorrelation length
  * Self test for `autocorr_acc`, which verifies the error bars and thus the autocorrelation length.

Fixes: #172


alea/include/alps/alea/internal/joined.hpp
```diff
@@ -18,12 +18,49 @@ namespace alps { namespace alea {
...
 }}
 
+namespace alps { namespace alea { namespace internal {
+    template <typename... Pack> struct add_scalar;
+    template <typename... Pack> struct joined_value;
+}}}
+
 // Actual declarations
 
 namespace alps { namespace alea { namespace internal {
```
egull: Please put comments abut what this does

alea/include/alps/alea/internal/joined.hpp
```diff
@@ -18,12 +18,49 @@ namespace alps { namespace alea {
 
     struct circular_var;
     struct elliptic_var;
+
 }}
```
egull: explanations?

alea/include/alps/alea/internal/pooling.hpp
```diff
@@ -0,0 +1,102 @@
...
+#include <alps/alea/covariance.hpp>
+#include <alps/alea/autocorr.hpp>
+
+#include <alps/alea/internal/joined.hpp>
+
+namespace alps { namespace alea { namespace internal {
+
+template <typename T, typename Str>
+const typename eigen<typename traits<cov_result<T,Str>>::cov_type>::matrix &
+get_cov(const cov_result<T,Str> &result)
```
egull: please... more documentation here.

alea/include/alps/alea/util.hpp
```diff
@@ -33,6 +37,9 @@ struct make_real { typedef T type; };
 template <typename T>
 struct make_real< std::complex<T> > { typedef T type; };
 
+template <typename T>
+using make_real_type = typename make_real<T>::type;
+
 template <typename T>
```
egull: again, more comments please

alea/test/model.cpp
```diff
@@ -0,0 +1,119 @@
...
+void fill(const alps::alea::util::var1_model<T> &model, Acc &acc, size_t tmax)
+{
+    alps::alea::util::var1_run<T> run = model.start();
+    boost::random::mt19937 engine;
+    while (run.t() < tmax) {
+        run.step(engine);
+        acc << run.xt();
+    }
+}
+
```
egull: this woudl benefit from 1/2 page of pedagogy

alea/include/alps/alea/autocorr.hpp
```diff
@@ -154,6 +154,9 @@ class autocorr_result
     /** Returns sample size, i.e., number of accumulated data points */
     size_t count() const { return level_[0].count(); }
 
+    /** Return sum of squared sizes */
```
galexv: Ehh, "squared sizes" of what?

alea/include/alps/alea/internal/joined.hpp
```diff
@@ -18,12 +18,49 @@ namespace alps { namespace alea {
...
 
+namespace alps { namespace alea { namespace internal {
```
galexv: Minor point: is `internal::` the same meaning as `detail::`? Should we use `detail::` for uniformity?
Or is it to follow some Eigen convention?

Basically, as far as i can understand, `detail::` (e.g., in Boost) means "implementation details, can be changed or disappear without warning, do not use!" (so to say, a "`private:` namespace"); while `internal::` (in Eigen) is expandable and Eigen [endorses it's use](https://eigen.tuxfamily.org/dox/TopicNewExpressionType.html) (so to say, a "`protected:` namespace"). 

Which one is the case here? Are non-alea developers allowed to add classes to `alps::alea::internal`? Why would they want to do it?

alea/include/alps/alea/autocorr.hpp
```diff
@@ -154,6 +154,9 @@ class autocorr_result
...
+    double count2() const { return level_[find_level(DEFAULT_MIN_SAMPLES)].count(); }
```
galexv: Not sure if `count2()` is a good name. Maybe, `count_squared()` or something? How important is it to keep the name short?

alea/include/alps/alea/internal/pooling.hpp
```diff
@@ -0,0 +1,102 @@
...
+template <typename Result1, typename Result2,
+          typename T=joined_value_type<Result1, Result2>>
+var_result<T> pool_var(const Result1 &r1, const Result2 &r2)
+{
+    if (r1.size() != r2.size())
+        throw size_mismatch();
+
+    var_result<T> pooled(var_data<T>(r1.size()));
+    pooled.store().count() = r1.count() * r2.count() / (r1.count() + r2.count());
+    pooled.store().count2() = pooled.store().count();   // FIXME
```
galexv: Why `FIXME`?

alea/include/alps/alea/internal/joined.hpp
mwallerb: It is a little bit of both, but closer to boost's `detail`, I think.  I would put it like this: it can change as long as the change does not break any real-world code :-)

I don't particularly like the name `detail` for a namespace, because it suggests that this is just a detail rather than a protected/private, i.e., internal, part of the interface.

alea/src/model.cpp
```diff
@@ -0,0 +1,131 @@
...
+    , t_(0)
+    , xt_(model.phi0())
+    , epst_(model.size())
+{  }
+
+template <typename T>
+void var1_run<T>::restart()
+{
+    t_ = 0;
+    xt_ = model_->phi0();
```
krivenko: How do you actually define the initial condition?
I can see two options: `X(0) = phi0` and `X(t) = phi0 + eps(0)`.

alea/include/alps/alea/internal/pooling.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+template <typename T, typename Str>
+const typename eigen<typename traits<cov_result<T,Str>>::cov_type>::matrix &
+get_cov(const cov_result<T,Str> &result)
+{
+    return result.cov();
+}
+
+template <typename Result>
+typename eigen<typename traits<Result>::cov_type>::matrix
```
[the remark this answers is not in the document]
mwallerb: > Why is the return type different from the previous overload (value vs const ref)?

For performance reasons. In the above case, the covariance is stored and can be directly returned, in the other it is constructed on-demand.  This is okay I think, since this is a very much internal method.

alea/include/alps/alea/internal/pooling.hpp
```diff
@@ -0,0 +1,102 @@
...
+        // Pooling covariance matrices - diagonalize those to yield variances
+        auto pooled_cov = r1.count() * get_cov(r1) + r2.count() * get_cov(r2)
+                          / (r1.count() + r2.count() - 2.0);
+
+        Eigen::SelfAdjointEigenSolver<typename eigen<T>::matrix> ecov(pooled_cov);
+        pooled.store().data() = ecov.eigenvectors().adjoint() * pooled.store().data();
+        pooled.store().data2() = ecov.eigenvalues();
+    } else {
+        // Directly pooling variances
+        pooled.store().data2() = r1.count() * r1.var() + r2.count() * r2.var()
```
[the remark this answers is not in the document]
mwallerb: > This expression corresponds to eq. (7) in the paper, right?
> Shouldn't you then be writing r1.count() - 1 and r2.count() - 1

Yes, thank you.

alea/include/alps/alea/util/model.hpp
```diff
@@ -0,0 +1,164 @@
...
+    /** Reference to underlying model */
+    const var1_model<T> &model() const { return *model_; }
+
+protected:
+    template <typename RandomEngine>
+    void get_noise(RandomEngine &engine)
+    {
+        using boost::random::normal_distribution;
+        for (size_t i = 0; i != epst_.size(); ++i) {
+            normal_distribution<var_type> dist(0, model_->stddev_eps()[i]);
```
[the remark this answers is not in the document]
mwallerb: > Consider storing a vector of distributions as a data member.

No.  boost's and std's `normal_distribution` do not precompute any values, so caching those objects adds complexity and a potential performance hit.

alea/src/model.cpp
```diff
@@ -0,0 +1,131 @@
...
+    auto eye = eigen<T>::matrix::Identity(size(), size());
+    return (eye - phi1_).colPivHouseholderQr().solve(phi0_);
+}
+
+template <typename T, typename DerivedA, typename DerivedB>
+typename eigen<T>::matrix kronecker(const Eigen::MatrixBase<DerivedA> &a,
+                                    const Eigen::MatrixBase<DerivedB> &b)
+{
+    typename eigen<T>::matrix res(a.rows() + b.rows(), a.cols() + b.cols());
+    for (int j = 0; j != a.cols(); ++j)
```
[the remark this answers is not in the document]
mwallerb: > size_t, for consistency

No.  `int` is the correct type here, since it is what Eigen uses as bounds.

alea/src/model.cpp
mwallerb: > How do you actually define the initial condition?
> I can see two options: X(0) = phi0 and X(t) = phi0 + eps(0).

Correct.  For simplicity, I went with the first option, though I agree the second is equally good.

alea/src/testing.cpp
```diff
@@ -0,0 +1,53 @@
...
+                  const column<typename make_real<T>::type> &var,
+                  double nmeas, size_t pools, double atol)
+{
+//     std::cerr << "\nDIFF= " << diff.transpose()
+//               << "\nVAR=  " << var.transpose()
+//               << "\nNMEAS=" << nmeas << "\n";
+
+    if (diff.rows() != var.rows())
+        throw std::invalid_argument("Size mismatch between diff and var");
+    if (pools != 1 && pools != 2)
```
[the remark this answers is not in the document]
mwallerb: > How about making pools a compile-time constant?

I don't see why. It just adds complexity and takes away flexibility.

alea/src/testing.cpp
```diff
@@ -0,0 +1,53 @@
+#include <alps/alea/testing.hpp>
+
+namespace alps { namespace alea {
+
+template <typename T>
+t2_result t2_test(const column<T> &diff,
+                  const column<typename make_real<T>::type> &var,
+                  double nmeas, size_t pools, double atol)
+{
+//     std::cerr << "\nDIFF= " << diff.transpose()
```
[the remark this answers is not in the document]
mwallerb: > Do you still need these lines?

I would like to keep them somehow, since they are really useful for debugging that cruft.

alea/src/testing.cpp
```diff
@@ -0,0 +1,53 @@
...
+//               << "\nNMEAS=" << nmeas << "\n";
+
+    if (diff.rows() != var.rows())
+        throw std::invalid_argument("Size mismatch between diff and var");
+    if (pools != 1 && pools != 2)
+        throw std::invalid_argument("Pools must be 1 or 2");
+    if (nmeas <= 0)
+        throw std::invalid_argument("Must have at least 1 measurement");
+    if ((var.array() < -atol).any())
+        throw std::invalid_argument("Variances must be positive");
```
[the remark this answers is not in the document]
mwallerb: > According to your if condition, the variances must not be too negative. Therefore, the error message is a bit misleading. And, btw, why not just compare to 0?

Because these variances may come from a diagonalization step, which means that they might be zero within a numerical error. That is what is written here.

alea/src/testing.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+    // remove data points where both the variance and difference are zero
+    // within atol while propagating intrinsic NaNs
+    double t2 = 0;
+    size_t nsize = 0;
+    for (int i = 0; i != diff.rows(); ++i) {
```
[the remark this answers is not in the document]
mwallerb: > size_t, for consistency.

Again, no.  Eigen indices are `int`s.

alea/test/model.cpp
```diff
@@ -0,0 +1,119 @@
...
+    alps::alea::autocorr_result<double> res2 = acc2.finalize();
+
+    std::cerr << "EST.  MEAN=" << res1.mean().transpose() << "\n";
+    std::cerr << "EST.  ERR =" << res1.stderror().transpose() << "\n";
+    std::cerr << "EST.  MEAN=" << res2.mean().transpose() << "\n";
+    std::cerr << "EST.  ERR =" << res2.stderror().transpose() << "\n";
+
+    // perform T2 test
+    alps::alea::t2_result t2 = alps::alea::test_mean(res1, res2);
+    ASSERT_GE(t2.pvalue(), 0.01);
```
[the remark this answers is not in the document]
mwallerb: > 

> This assertion fails on my laptop
> /home/igor/Physics/ALPSCore/ALPSCore.git/alea/test/model.cpp:118: Failure
Expected: (t2.pvalue()) >= (0.01), actual: 2.60007e-06 vs 0.01
> Might the failure have something to do with boost::random::mt19937 that is used in fill()?
Boost PRNGs and distributions are known to give different sequences in different versions (cf. TRIQS/triqs#226).

That is curious. I messed up something there ...

alea/include/alps/alea/autocorr.hpp
mwallerb: I neither find `count2` or `count_squared` to be particularly descriptive. (It's the sum of squared weights).  For consistency with `data2`, I therefore named it `count2`.

galexv:
The [`model` test is failing](https://travis-ci.org/ALPSCore/ALPSCore/jobs/376419313#L3546). Is it expected?

egull:
@mwallerb could you please have another look at this?

galexv:
@mwallerb @egull 
This PR still has a failing test `model`:
```diff
cd ALPSCore/build/alea/test && ulimit -d 2500000 && make -j3 model && ./model
[...]
Running main() from gtest_main.cc
[==========] Running 5 tests from 2 test cases.
 [----------] Global test environment set-up.
 [----------] 3 tests from var1_test_model
 [ RUN      ] var1_test_model.mean
PHI0=
2
3
PHI1=
 0.5 0.25
0.25  0.3
VEPS=
   1
0.25
+[       OK ] var1_test_model.mean (1 ms)
[ RUN      ] var1_test_model.cov
PHI0=
2
3
PHI1=
 0.5 0.25
0.25  0.3
VEPS=
   1
0.25
+[       OK ] var1_test_model.cov (1 ms)
[ RUN      ] var1_test_model.tau
PHI0=
2
3
PHI1=
 0.5 0.25
0.25  0.3
VEPS=
   1
0.25
+[       OK ] var1_test_model.tau (0 ms)
[----------] 3 tests from var1_test_model (2 ms total)

[----------] 2 tests from var1_test
[ RUN      ] var1_test.autocorr
EXACT MEAN=     10 8.33333
EXACT TAU =      4 1.77778
EST.  MEAN= 10.006 8.33161
EST.  ERR =0.00771677 0.00220284
EST.  TAU = 3.8157 1.79048
-alea/test/model.cpp:90: Failure
-Expected: (t2.pvalue()) >= (0.01), actual: 0.00238033 vs 0.01
-[  FAILED  ] var1_test.autocorr (2675 ms)
[ RUN      ] var1_test.same
EST.  MEAN=20.0118 4.28484
EST.  ERR =0.0152815 0.0011353
EST.  MEAN=20.0169 4.28255
EST.  ERR = 0.0216039 0.00409343
-alea/test/model.cpp:119: Failure
-Expected: (t2.pvalue()) >= (0.01), actual: 2.60007e-06 vs 0.01
-[  FAILED  ] var1_test.same (5353 ms)
[----------] 2 tests from var1_test (8028 ms total)

[----------] Global test environment tear-down
[==========] 5 tests from 2 test cases ran. (8032 ms total)
+[  PASSED  ] 3 tests.
-[  FAILED  ] 2 tests, listed below:
-[  FAILED  ] var1_test.autocorr
-[  FAILED  ] var1_test.same

-2 FAILED TESTS
```
I am **not** includng this PR into today's release.

alea/test/model.cpp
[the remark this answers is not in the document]
mwallerb: Should work now

alea/include/alps/alea/internal/joined.hpp
mwallerb: These are only forward declarations, the documentation is with the actual declaration.

alea/include/alps/alea/internal/joined.hpp
mwallerb: These are only forward declarations, the documentation is with the actual declaration.

mwallerb:
Fixed the tests and some code.  Propose to merge.

PR #491, opened 2018-03-21
Removed some more warnings

galexv:
(Ongoing issue #375)

### DIFF
--- a/gf/test/gf_new_tail_test_mpi.cpp
+++ b/gf/test/gf_new_tail_test_mpi.cpp
@@ -34,7 +34,7 @@ public:
   omega_sigma_gf gf2;
   typedef omega_sigma_gf gf_type;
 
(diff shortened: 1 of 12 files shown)


gf/test/gf_new_tail_test_mpi.cpp
```diff
@@ -34,7 +34,7 @@ class TailedGreensFunctionMPI : public ::testing::Test
   omega_sigma_gf gf2;
   typedef omega_sigma_gf gf_type;
 
-  TailedGreensFunctionMPI():beta(10), nfreq(10), nsites(5), nspins(2),
+  TailedGreensFunctionMPI():beta(10), nsites(5), nfreq(10), nspins(2),
```
egull: what's the rationale of this change? (not that it does anything...)

galexv: It eliminates warning about different order of declarations in the class and initializers in the ctor.

PR #499, opened 2018-04-03
C++14 in Travis CI

shinaoka:
We build ALPSCore with C++14 in Travis CI.
As expected, builds with C++14 fail due to the bug in Boost 1.58.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,10 +22,16 @@ addons:
 install: true
 
 env:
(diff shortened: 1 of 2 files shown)


shinaoka:
@aeantipov 

This PR addresses the following issues.

- Introduced tests for (clang, gcc)  x (C++11, C++14)
- A workaround for a bug in Boost 1.58.0 (C++14)

Could you review this?

Issue #484, opened 2018-03-01
[ALEA] Return types of count() and count2() are different between *_data and *_result classes


krivenko:
Return types of count() and count2() methods are inconsistent (`size_t` vs `double`) between result classes and their respective data classes. Out of 5 pairs of classes, only `mean_result`/`mean_data` aren't affected by this problem, I guess.

krivenko:
Wait a minute, `count2()` still has inconsistent return types...
```
(13:00) igor@floquet ~/Physics/ALPSCore/ALPSCore.git/alea/include/alps/alea [pr507] $ grep -R "count2() const" *
batch.hpp:    double count2() const { return store_->count().squaredNorm(); }
covariance.hpp:    double count2() const { return count2_; }
covariance.hpp:    size_t count2() const { return store_->count2(); }
variance.hpp:    double count2() const { return count2_; }
variance.hpp:    size_t count2() const { return store_->count2(); }
```

mwallerb:
Yes, but this needs to be that way, because `size2_` is potentially `size_ * size_`

mwallerb:
Ah no, shit, you're right ...

PR #507, opened 2018-05-09
[alea] Make `count()` consistent over accumulators.

mwallerb:
Also, adds some tests to the generic (variant-based) result class to match that.

Closes: #484

### DIFF
--- a/alea/test/result.cpp
+++ b/alea/test/result.cpp
@@ -1,0 +1,94 @@
+#include <alps/alea.hpp>
+#include <alps/testing/near.hpp>
+
+#include "gtest/gtest.h"
+
+template <typename Acc>
+class result_case
+    : public ::testing::Test
+{
+public:
+    typedef Acc acc_type;
+    typedef alps::alea::traits<Acc> acc_traits;
+    typedef typename acc_traits::store_type  store_type;
+    typedef typename acc_traits::value_type value_type;
+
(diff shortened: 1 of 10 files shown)


alea/test/result.cpp
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    void test_result_caps()
+    {
+        EXPECT_EQ(res.size(), generic_res.size());
+        EXPECT_EQ(res.count(), generic_res.count());
+        ALPS_EXPECT_NEAR(res.mean(), generic_res.mean<value_type>(), 1e-16);
+    }
+
+    // Thank you, C++, for this very nice way of doing "if/else".  Go die.
```
krivenko: Too bad, we cannot have `constexpr if` just yet.
Here is a little bit less annoying way of doing this.
```c++
    // Have VAR
    void test_var(std::true_type) {
        typedef typename result_traits::value_type value_type;
        ALPS_EXPECT_NEAR(res.var(), generic_res.var<value_type>(), 1e-16);
    }

    // Does not have VAR
    void test_var(std::false_type) {
        typedef typename result_traits::value_type value_type;
        EXPECT_THROW(generic_res.var<value_type>(), alps::alea::estimate_unavailable);
    }

...

TYPED_TEST(result_case, test_var) {
    this->test_var(std::integral_constant<bool, TestFixture::result_traits::HAVE_VAR>());
}
```

Issue #506, opened 2018-04-26
[alea] linear_prop is broken for batched accumulators


kleinhenz:
Linear propagation gives incorrect results when the input accumulator is batched, e.g. `batch_acc` or `cov_acc` with `bundle_size > 1`. I think this has to do with `count2` not being computed correctly.

MWE

```
#include <iostream>
#include <random>
#include <vector>
#include <alps/alea.hpp>

int main(int argc, char* argv[]){
  std::random_device rd{};
  std::mt19937 gen {rd()};

  std::normal_distribution<> d {0.0, 2.0};
  alps::alea::batch_acc<double> acc(1);

  int N = 1000000;
  for (int i = 0; i < N; ++i) {
    acc << d(gen);
  }

  auto result = acc.result();

  std::function<double(double)> f = [](double x) { return x; };
  auto tf = alps::alea::make_transformer(f);
  auto transformed_result =
      alps::alea::transform(alps::alea::linear_prop(), tf, result);

  std::cout << result << std::endl;
  std::cout << transformed_result << std::endl;
}
```

Produces output

```
0.00108921 +- 0.00194441
0.00108921 +- 0.123001
```

krivenko:
The problem is here:
https://github.com/ALPSCore/ALPSCore/blob/e70f3a9820e41276e9825f8da080419782f67b93/alea/include/alps/alea/transform.hpp#L62

If `in` is a `cov_result`, `in.cov()` returns `batch_size() * store_->data2()`. I guess one has to rescale the result back before storing it in `res.store().data2()`.

egull:
@mwallerb please make fixing this a priority.

PR #512, opened 2018-05-09
[alea] properly rescale (co-)variances before applying transform.

mwallerb:
Also adds a test.

This issue is the side product of the fact that the variance of
a set of weighted measurements and the "ensemble" variance differ
by a factor, so we have to correct for that factor in the computation
of the variance, which means we have to undo that before we
transform that variance.  The clean solution would be to have a
"batching decorator" that makes this transformation.

Fixes: #506


alea/test/twogauss.cpp
```diff
@@ -88,6 +88,13 @@ class twogauss_mean_case
         test_result();    // keeps mean constant
     }
 
+    void test_resize()
```
krivenko: You still need to call this function.

mwallerb: Fixed.

Issue #511, opened 2018-05-09
[travis] run tests in parallel to check for MPI issues


mwallerb:
Given bugs #508 and #509, I think we should run tests in parallel on travis.

Something like

    $ ALPS_TEST_MPI_NPROC=4 make test

here: https://github.com/ALPSCore/ALPSCore/blob/master/.travis.yml#L59

galexv:
Closed by commit c1c69da06137b7c01f1f73379fec03d092ff8f0c.

Issue #509, opened 2018-05-09
[gf] test `mesh_test_mpi` fails when run in parallel


mwallerb:
When running:

    $ mpirun -np 2 ./gf/test/mesh_test_mpi

I get:

    [ RUN      ] MeshTest.MpiBcastMatsubara
    /home/mwallerb/Workspace/alpscore/gf/test/mesh_test_mpi.cpp:54: Failure
    Value of: ref_mesh
      Actual: # MATSUBARA mesh: N: 20 beta: 5 statistics: FERMIONIC POSITIVE_NEGATIVE
    Expected: *mesh_ptr
    Which is: # MATSUBARA mesh: N: 20 beta: 5 statistics: FERMIONIC POSITIVE_NEGATIVE
    Failed at rank=1
    [  FAILED  ] MeshTest.MpiBcastMatsubara (0 ms)

NB: I think we should run the tests in parallel on travis as well to check for that sort of thing.

mwallerb:
Fixed by #516, right?

Issue #508, opened 2018-05-09
[accumulators] test `repeated_merge` fails when run in parallel


mwallerb:
On the old accumulator framework, when running:

    $ mpirun -np 2 ./accumulators/test/repeated_merge

I get:

    /home/mwallerb/Workspace/alpscore/accumulators/test/repeated_merge.cpp:91: Failure
    Value of: rset["value"].count()
      Actual: 50000
    Expected: rset["test_value"].count()
    Which is: 40000
    Count
    [  FAILED  ] AccumulatorTest/1.twoMerges, where 
      TypeParam = (anonymous namespace)::proxy<alps::accumulators::NoBinningAccumulator, double> (20 ms)

It fails for all other accumulator types as well.

NB: I think we should run the tests in parallel on travis as well to check for that sort of thing.

galexv:
Looking at it...

galexv:
Commit c987a08cf5b18d54149e164aecbb0d9feed67bec is last good (aka PR #380), commit 97d4b696ca4ec0ace8090fbf7aa4b7f9837ca0d1 is first bad: merging the master into the feature branch (issue #295).

PR #514, opened 2018-05-10
Run MPI tests in parallel on Travis

galexv:
...as suggested by Markus (@mwallerb), see issue #511.


galexv:
The TravisCI tests are failing as expected, due to issues #508 (PR #513 pending) and #509. Once those are fixed, the tests should pass.

egull:
A bit ugly that the non-MPI tests have an MPI specific variable set, too. But I guess that's ok. No need to overthink it.

galexv:
> A bit ugly that the non-MPI tests have an MPI specific variable set, too. But I guess that's ok.

I think so too: non-MPI builds simply do not use them, so it's perfectly ok. Otherwise scripts could become hairy: either if-conditions or conditionals substitutions of variables.

PR #516, opened 2018-05-12
Update mesh.hpp

iskakoff:
#509


egull:
Looks ok to me. @galexv please have a quick look then merge first thing...

egull:
So what I really don't get is why this had to be run with MPI to trigger the exception... ??

iskakoff:
@egull Because it's trying to broadcast a mesh into another mesh, you can't do it on a single core in principle.

PR #521, opened 2018-07-02
Try to download Eigen3 using various methods

### LINKED ISSUE #451: ALPSCore cannot download Eigen3
...if CMake is compiled without SSL support.

Then user has to download Eigen3 themselves (then *clear the cache* and use `-DALPS_EIGEN_TGZ_FILE=$PWD/3.3.4.tar.gz`) .

We should implement workarounds: try git, try hg, try wget, try curl.

galexv:
The user can set `-DALPS_EIGEN_DOWNLOAD_METHODS` to a list of download methods to try, in the given order. Available methods are `cmake` (using CMake built-in `file`), `wget` (using `wget` utility, if available) and `curl` (using `curl` utility, if available). The default is `cmake;wget;curl`.

This should close #451.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,7 @@ script:
     -DCMAKE_CXX_COMPILER=${ALPS_CXX:-${CXX}}              \
     -DCMAKE_INSTALL_PREFIX=$TRAVIS_BUILD_DIR/installed    \
     -DALPS_INSTALL_EIGEN=true                             \
-    -DBUNDLE_DOWNLOAD_TRIES=3                             \
+    -DALPS_BUNDLE_DOWNLOAD_TRIES=3                        \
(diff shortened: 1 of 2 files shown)


egull:
Looks good, as far as I can see. One addition: here,

            message(FATAL_ERROR "Failed to download ${ALPS_EIGEN_TGZ_FILE} "
             "from ${ALPS_EIGEN_DOWNLOAD_LOCATION}: "
             "status=" ${status_}) 

it would be useful to say something like 'please download eigen manually and put it to the following location...' or 'please download eigen manually, unzip it, and specify its location as ...'
whatever makes sense. That way users know what to do if it doesn't work.

galexv:
I expanded the error messages and added more safeguards.

PR #532, opened 2018-07-18
User-settable linear function for RNG seeds in parallel MC

### LINKED ISSUE #307: Random seed in Monte Carlo simulations
I got some feedback from a user.
He ran our ALPSCore/CT-HYB several times by increasing the random seed one by one to estimate error values of Green's function in MPI calculations.
This will lead to a wrong estimate of error vars because we use the given "SEED" + I for process I.
https://github.com/ALPSCore/ALPSCore/blob/master/mc/include/alps/mc/mpiadapter.hpp#L37

Should we do something like this? 
   Base(parameters, 7*comm.size()*comm.rank())

Or, should we document it?

galexv:
RNG seeds of MC chains depend linearly on the chain MPI rank; the
coefficients of the linear dependence can now be set by the caller.

@shinaoka Could you please also have a brief look if you have time?

This should close #307.

### DIFF
--- a/mc/include/alps/mc/mpiadapter.hpp
+++ b/mc/include/alps/mc/mpiadapter.hpp
@@ -29,12 +29,26 @@ namespace alps {
 
         protected:
             /// Construct mcmpiadapter_base with a custom scheduler
+            /**
+               Initializes the wrapped simulation class, passes parameters and an RNG seed that linearly depends on MPI rank.
+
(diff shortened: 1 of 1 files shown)


galexv:
It lacks a proper unit test, but for this we can open a separate issue.

egull:
Let's wait for input from Hiroshi before closing.

PR #537, opened 2018-08-03
Update for gf2

iskakoff:


utilities/include/alps/numeric/tensors/data_storage.hpp
```diff
@@ -7,11 +7,43 @@
...
+         * release memory
+         * @param data
+         */
+        void deallocate(T* data) {
+          delete [] data;
+        }
+
+        void lock(){}
+        void release(){}
+        bool locked() {return true;}
```
mwallerb: These methods are the central ingredient to your home-brew allocator protocol, so they are in dire need of documentation.  In particular, it is very surprising that locks are bound to the allocator, rather than the allocated memory, and this can only be explained by the fact that this is an abstraction of the MPI model.

utilities/include/alps/numeric/tensors/mpi/mpi_shared_allocator.hpp
```diff
@@ -0,0 +1,86 @@
...
+          MPI_Comm_size(*comm_, &s);
+          MPI_Comm_rank(*comm_, &r);
+          size_t nloc = n / s;
+          if(n%s > r) nloc++;
+          return nloc;
+        }
+
+        MPI_Comm *comm_;
+        MPI_Win *win_;
+        bool lock_;
```
mwallerb: The undressed pointers combined with a cached flag `lock_` means this class is *very* unsafe to copy.

utilities/include/alps/numeric/tensors/mpi/mpi_shared_allocator.hpp
```diff
@@ -0,0 +1,86 @@
...
+        }
+        void lock(){
+          lock_ = true;
+          MPI_Win_lock_all(MPI_MODE_NOCHECK , *win_);
+        }
+        void release(){
+          MPI_Win_sync(*win_);
+          MPI_Win_unlock_all(*win_);
+          lock_ = false;
+        }
```
mwallerb: Hang on, isn't that allocator locking model broken by design?

Suppose you have two tensors with two shared_allocators, say `a` and `b`, so you will get two allocators where the locks are bound to the same window, in other words, if  allocator `b` locks, then `b.lock_ = false`, even though the underlying window is locked, isn't it?

utilities/include/alps/numeric/tensors/mpi/mpi_shared_allocator.hpp
```diff
@@ -0,0 +1,86 @@
...
+        mpi_shared_allocator(MPI_Comm& comm, MPI_Win& win) : comm_(&comm), win_(&win) {
+
+        };
+
+        T* allocate(size_t size) {
+          void *mem;
+          // compute local size for array
+          size_t nloc = local_size(size);
+          // allocate shared memory
+          MPI_Win_allocate_shared(nloc * sizeof(T), sizeof(T), MPI_INFO_NULL, *comm_, &mem, win_);
```
mwallerb: Also, here you override the stored window in `win_`, so why do you need the parameter in the constructor?

utilities/include/alps/numeric/tensors/mpi/mpi_shared_allocator.hpp
```diff
@@ -0,0 +1,86 @@
...
+         * @return local size
+         */
+        size_t local_size(size_t n) const {
+          int s, r;
+          MPI_Comm_size(*comm_, &s);
+          MPI_Comm_rank(*comm_, &r);
+          size_t nloc = n / s;
+          if(n%s > r) nloc++;
+          return nloc;
+        }
```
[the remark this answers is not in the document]
iskakoff: Disagree. The way to distribute data between the cpus are task-specific.

utilities/include/alps/numeric/tensors/data_storage.hpp
```diff
@@ -20,45 +52,154 @@ namespace alps {
...
+        /// size of buffer
+        size_t size_;
+        /// dataspace allocator
+        Allocator allocator_;
+        /// Check the data is locked in case of shared memory access
+        void check_lock() {
+          if(!allocator_.locked()) {
+            std::cerr<<"You are trying to access non-locked tensor with non-const method."<<std::endl;
+          }
+        }
```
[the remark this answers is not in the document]
iskakoff: No way I will put throw there. If your object is non-const, it will always use non-const function. so you will always then have non-const access even if you are trying to read data.

utilities/include/alps/numeric/tensors/tensor_base.hpp
```diff
@@ -32,9 +32,22 @@ namespace alps {
...
       };
 
+      /**
+       * Check that all values in pack are true
+       *
+       * @tparam IndicesTypes - types of indices template parameter pack
+       */
+      template <bool...> struct bool_pack;
+      template <bool... v>
+      using all_true = std::is_same<bool_pack<true, v...>, bool_pack<v..., true>>;
```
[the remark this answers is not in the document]
iskakoff: How is it better?

utilities/include/alps/numeric/tensors/tensor_base.hpp
```diff
@@ -108,22 +133,42 @@ namespace alps {
          */
         tensor_base(Container &&container, const std::array < size_t, Dim >& sizes) : storage_(container), shape_(sizes) {
           static_assert(is_storage< T, Container>::value, "Should be either data_storage or data_view type");
+          assert(storage_.size() == size());
```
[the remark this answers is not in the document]
iskakoff: Because you need to know what are you doing. Having additional condition for release mode is slow. 

You have the same in Eigen, it checks  boundaries, correct dimensions and so on  only in debug mode.

gf/include/alps/gf/gf_base.hpp
```diff
@@ -174,7 +174,11 @@ namespace alps {
         /// Create GF with the provided data
         gf_base(VTYPE* data, const mesh_types &meshes) : data_(data, get_sizes(meshes)), meshes_(meshes), empty_(false) {}
         /// Create GF with the provided data and meshes
-        gf_base(data_storage const &data, MESHES...meshes) : data_(data), meshes_(std::make_tuple(meshes...)), empty_(false) {}
+        gf_base(const data_storage &data, const mesh_types &meshes) : data_(data), meshes_(meshes), empty_(false) {}
+        /// Create GF with the provided data and meshes
+        gf_base(data_storage &&data, const mesh_types &meshes) : data_(data), meshes_(meshes), empty_(false) {}
+        /// Create GF with the provided data and meshes
+        gf_base(const data_storage &data, MESHES...meshes) : data_(data), meshes_(std::make_tuple(meshes...)), empty_(false) {}
```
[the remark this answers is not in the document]
iskakoff: This part came from tails, tails came from HDF5, HDF5 came from tensors. So, all these changes are related.

utilities/include/alps/numeric/tensors/data_storage.hpp
```diff
@@ -7,11 +7,43 @@
...
 namespace alps {
   namespace numerics {
     namespace detail {
 
+      /**
+       * Simple data allocator
+       * @tparam T - datatype
+       */
+      template <typename T>
+      class simple_allocator {
```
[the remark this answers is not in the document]
iskakoff: because I need additional semantic for locks

utilities/include/alps/numeric/tensors/data_storage.hpp
iskakoff: Because locks are connected with MPI window.

utilities/include/alps/numeric/tensors/mpi/mpi_shared_allocator.hpp
iskakoff: In other word. When you trying to use MPI, you should think properly, and understand what you are doing. 

Saying "broken by design" you can say simply that MPI is broken by design. Which is not, you just need to properly design your algorithm.

utilities/include/alps/numeric/tensors/mpi/mpi_shared_allocator.hpp
iskakoff: I agree that the shared memory semantic should be revised, since it was just scratch implementation and I never use it in any codes, and I will probably for now remove it from this PR since I am too lazy to think about it right now.

The idea was the following. 
1. Allocation and deallocation should be done through MPI window.  This is the only collective operations.
2. When you access data for writing you should lock window
3. Reading can be done as simple "operator[]" access
4. When you copy tensor you can't copy data, since it's shared-allocated and bound to window, so, you need to copy pointer to the underlying data (this is definitely not implemented), and copy pointer to the window. 
5. When you destroy the tensor, your memory should not be released, since it is still connected to the window. And you still get access through MPI_Get, MPI_Put methods.
6. When the shared memory region is not needed anymore you should free this window (see 1.)

utilities/include/alps/numeric/tensors/data_storage.hpp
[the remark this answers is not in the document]
mwallerb: Understood.  I'm saying you should update the docs to that effect.

utilities/include/alps/numeric/tensors/data_storage.hpp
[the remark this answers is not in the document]
mwallerb: Ah okay, didn't think of that!

utilities/include/alps/numeric/tensors/mpi/mpi_shared_allocator.hpp
mwallerb: No, I'm sorry but you cannot blame this on MPI alone.  This class is just not well-designed.  This code breaks:

```C++
  mpi_shared_allocator a(...);
  mpi_shared_allocator b = a;
  b.lock();
  // a is not locked; i.e., use a -> boom.
```

which also means you cannot embed the allocator into a class, which you can usually do:

```C++
template <T, Alloc>
class my_vector {

private:
     T *begin, *end, *capacity;
     Alloc something;  // put in mpi_shared_allocator -> boom.
```

Allocations leak memory:

```C++
  {
    MPI_Win win;
    ...
    mpi_shared_allocator a(win);
    // gets new window which you dont have access to
    T *p1 = a.allocate(10);
    a.deallocate(p1);
  }  // a is destructor -> leaks internal window -> boom.
```

gf/include/alps/gf/gf_base.hpp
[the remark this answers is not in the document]
mwallerb: Ah okay, fine then, forgot about the interdependencies.

utilities/include/alps/numeric/tensors/mpi/mpi_shared_allocator.hpp
mwallerb: That seems very reasonable.  As far as I understand it, a shared memory is always tied to a window, correct?  In this case, how about bundling the two together, IOW, the allocator returns something like this:

```C++
template <typename T> struct mpi_shared_memory {
      MPI_Win window;
      T *memory;
 };
```

So in this case you can make sure that you always destroy the window with it.  Would that work?

utilities/include/alps/numeric/tensors/data_storage.hpp
mwallerb: I understand.  But I think it should have some documentation to that effect.

utilities/include/alps/numeric/tensors/mpi/mpi_shared_allocator.hpp
iskakoff:
```C++
  mpi_shared_allocator a(...);
  mpi_shared_allocator b = a;
  b.lock();
  // a is not locked; i.e., use a -> boom.
```

as long as you read it - there is nothing critical. You wil probably not get the most recent RMA updates since your cached data is wrong. But this is again, I can do exactly the same with pure MPI without any encapsulation.

> which also means you cannot embed the allocator into a class, which you can usually do:

```C++
template <T, Alloc>
class my_vector {

private:
     T *begin, *end, *capacity;
     Alloc something;  // put in mpi_shared_allocator -> boom.
```

You still can do this if you do proper copy-construction of allocator.

> Allocations leak memory:

no they don't, you should manage memory with window. That is the reason to have window outside. It's the way to have proper copy-semantic. You can't do any copy-construction of shared object If you free memory within allocator destructor. Other way is to disallow copying of shared allocator, so you can't assign or pass by value shared object. 

I don't know which one is better. The second seems easy to implement, but it's kinda not flexible.

BTW, I will remove shared-memory from this PR, don't have time for it right now.

mwallerb: Again, either I am really, really missing something, or I think it leaks memory:

```C++
    MPI_Win *win = nullptr;
    {
        mpi_shared_allocator alloc(MPI_COMM_WORLD, win);

        // what happens in allocate():  you call `MPI_Win_allocate_shared`, which creates a
        // new window and overrides the `win_` variable with it. (you are still fine because
        // you have a handle to the old win_ from outside.)
        T *p1 = alloc.allocate(10);
        alloc.deallocate(p1);

        // Now you're screwed.  You just overrode the window with a new window without
        // calling MPI_Free_Win on the current one first.  This will usually happen 
        // somewhere deep in your code (e.g., a vector allocating space for more elements).
        T *p2 = alloc.allocate(10);
        alloc.deallocate(p2);
    }

    // At this point you destroy the allocator, but you DO NOT call MPI_Free_Win on the
    // first one which means the new internal window does not get destroyed.  You 
    // cannot destroy it yourself, because alloc does not give you access to it.
    if (win != nullptr)
        MPI_Free_Win(win);
```

mwallerb: Here is another, maybe less contrived, example:

```C++
struct mydata {
    mydata(mpi_shared_allocator<int> alloc) : alloc(alloc), data(alloc.allocate(10)) { }
    mydata(const mydataset &other) : mydataset(other) { }
    ~mydata() { alloc.deallocate(data); }    
private:
    mydata &operator=(const mydata &other) = delete;
    int *data;
    mpi_shared_allocator<int> alloc;
};

MPI_Win *win = nullptr;
mydata x(mpi_shared_allocator<int>(win));
mydata y = x;   // Boom: leaks old window, no way of fixing
if (win != nullptr)
        MPI_Win_free(*win);
```

iskakoff: Hey-hey-hey, careful.

You can not do this:

```C++
    MPI_Win *win = nullptr;
```

it should be

```C++
    MPI_Win win;
```

and then something like this.

```C++
    if (win != MPI_WIN_NULL)
        MPI_Win_free(&win);
```

mwallerb: OK, let's agree to disagree.  If we can't agree on whether those two examples point to an API that is fundamentally broken versus just one where one needs to understand what one is doing, we won't reach common ground here.

PR #542, opened 2018-08-08
Tensors refactoring

iskakoff:
1. change slice creation (avoid additional view construction)
2. fix bug with view sizes
3. introduce custom data container in storage (for future customization with shared memory allocation)
4. negate operator
5. assignment to constant scalar
6. some const-correctness
7. additional tests

### DIFF
--- a/utilities/include/alps/numeric/tensors/data_storage.hpp
+++ b/utilities/include/alps/numeric/tensors/data_storage.hpp
@@ -131,10 +131,6 @@ namespace alps {
     }
     template<typename T>
     using simple_storage = detail::data_storage<T, std::vector<T> >;
-#ifdef ALPS_HAVE_SHARED_ALLOCATOR
-//    template<typename T>
-//    using shared_storage = detail::data_storage<T, detail::mpi_shared_allocator<T> >;
-#endif
   }
 }
 
--- a/utilities/include/alps/numeric/tensors/data_storage.hpp
+++ b/utilities/include/alps/numeric/tensors/data_storage.hpp
@@ -7,6 +7,7 @@
 #ifndef ALPSCORE_GF_TENSORBASE_H
 #define ALPSCORE_GF_TENSORBASE_H
 
+#include <vector>
 
 namespace alps {
   namespace numerics {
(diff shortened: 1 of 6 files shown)


utilities/include/alps/numeric/tensors/data_storage.hpp
```diff
@@ -20,45 +21,94 @@ namespace alps {
        * @brief TensorBase class implements basic multi-dimensional operations.
        * All arithmetic operations are performed by Eigen library
        */
```
mwallerb: Now that you have two template parameters, it might be good to briefly explain what they do. Something to the effect of

```C++
  /**
   * @tparam T  the scalar type
   * @tparam Cont  abstraction of the data storage (default vector)
   */
```

utilities/include/alps/numeric/tensors/data_storage.hpp
```diff
@@ -20,45 +21,94 @@ namespace alps {
...
+          if(size() != rhs.size()) {
+            resize(rhs.size());
+          }
+          std::copy(rhs.data(), rhs.data() + rhs.size(), data());
+          return *this;
+        };
+
+        template<typename T2, typename C2>
+        data_storage<T, Cont>& operator=(const data_storage<T2, C2>& rhs) {
+          static_assert(std::is_convertible<T2, T>::value, "Can't perform assignment: T2 can be casted into T");
```
mwallerb: Typos: "can" -> "cannot"; also the past tense of "cast" is "cast" :)

utilities/include/alps/numeric/tensors/tensor_base.hpp
```diff
@@ -301,8 +349,18 @@ namespace alps {
...
-          Eigen::Map < Eigen::Matrix < T, 1, Eigen::Dynamic > > M(&storage_.data(0), storage_.size());
-          M.setZero();
+          set_number(T(0));
+        }
+
+        /**
+         *
+         * @tparam num_type
+         * @param value
+         */
```
mwallerb: The doxygen comment here is a stub.

utilities/include/alps/numeric/tensors/tensor_base.hpp
```diff
@@ -316,6 +374,14 @@ namespace alps {
           return storage_.size();
         }
 
+        /**
+         * Negate tensor
+         */
+        tensor_base < T, Dim, Container > operator-() {
+          *this *= -1.0;
+          return *this;
+        };
```
mwallerb: This is strange behaviour for a unary minus. It modifies `this` in-place and then returns a copy of it.

Usually, it is a `const` method and returns a new object containing `-a`.  In other words, `-a` should behave like `0-a`.

Once you're at it, there should be a test for this I think.

PR #556, opened 2018-08-30
Striving for warning-free compilation

galexv:
Enable more warnings, and treat warnings as errors when building on TravisCI.
This is related to issue #375 .

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -2,20 +2,20 @@
     "license": "other-open", 
     "creators": [
         {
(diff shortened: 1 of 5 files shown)


galexv:
> I fear that we cause more trouble with the mpi hack than we solve. 

Well, the hack is confined to the TravisCI build script, and is specific to the TravisCI build environment. I think it should not affect the library users. Once Travis is updated and/or we have a controlled test environment, we can remove it.

PR #559, opened 2018-09-10
Override params restored from HDF5 via cmdline

### LINKED ISSUE #436: Overriding parameters when restarting from clone?
Hi,
I just updated to the latest ALPSCore master and was surprised to find that it is apparently no longer supported to override parameters via command line arguments when restoring from an hdf5 archive? As far as I can tell, the relevant changes have been introduced [here](https://github.com/ALPSCore/ALPSCore/commit/b1d48f373882d084d32f31466e611aeba10facfd#diff-3276e6c2e18a491d51b5cc05704fa614R105). I always found that to be a particularly useful feature, especially when overriding the `timelimit` or parameters which are related to post-processing. Is this feature gone for good, or would it be possible to reintroduce it again? Any technical reasons that warranted removal?

galexv:
This should close #436.

### DIFF
--- a/params/include/alps/params.hpp
+++ b/params/include/alps/params.hpp
@@ -143,6 +143,45 @@ namespace alps {
                 @param argv Array of pointers to command line arguments (as in `main(int argc, char** argv)`)
                 @param hdf5_path path to HDF5 dataset containing the saved parameter object
                        (NULL if this functionality is not needed)
+
+
+                The parameters are supplied in the `--key=value`
+                format; the double dash (`--`) can be shortened to a single dash (`-`)
+                or omitted.  The `=value` part may be omitted, in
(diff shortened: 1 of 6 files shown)


galexv:
Not quite ready. :(
For example:
```
$ ./mytest params1.ini params.h5 params2.ini 
Params were restored from archive params1.ini
Origin name: params1.ini
```
while the correct output should have been `Params were restored from archive params.h5`.
I will make a test for it and fix the issue within the same PR.

galexv:
@egull Emanuel, could you please have a look again? (Fixed https://github.com/ALPSCore/ALPSCore/pull/559#issuecomment-421462682)

egull:
yep, looks good and tests pass.

PR #568, opened 2018-10-08
Fix Eigen3 co-installation when requested as the 2nd attempt

### LINKED ISSUE #560: [compilation] Cannot co-install Eigen3 after failed CMake
Suppose a user tries to build ALPSCore:
```sh
$ cmake $alpscore_src 
```
and it fails because Eigen3 is not available:
```diff
 -- eigen requested
 -- Could NOT find Eigen3 (missing: EIGEN3_INCLUDE_DIR EIGEN3_VERSION_OK) (Required is at least version "3.3.4")
 CMake Error at common/cmake/ALPSEnableEigen.cmake:151 (message):
 [...]
-    3. Rerun CMake with option:
-       -DALPS_INSTALL_EIGEN=true 
-      to request the installation script to attempt to download Eigen3
-      and co-install it with ALPSCore.
 [...]
```
Now the user does as requested:
```sh
$ cmake $alpscore_src  -DALPS_INSTALL_EIGEN=true
```
and gets:
```diff
 -- Unpacked successfully
 -- Searching for Eigen3 in 
--- Could NOT find Eigen3 (missing: EIGEN3_INCLUDE_DIR EIGEN3_VERSION_OK) (Required is at least version "3.3.4")
-CMake Error at common/cmake/ALPSEnableEigen.cmake:240 (message):

-  Cannot find suitable Eigen3 in .  Make sure that Eigen3 is indeed at the
-  specified location, or unset the variable `EIGEN3_INCLUDE_DIR` to download
-  Eigen3 to your machine.
Call Stack (most recent call first):
  utilities/CMakeLists.txt:39 (add_eigen)
```

which _looks_ like a bug, and is surely confusing.
Removing the `CMakeCache.txt` helps, but we should either explicitly advise to do so, or take care of this situation ourselves (in this case, the supposed location of Eigen3 is empty; CMake script should ignore it, then!).

galexv:
This closes #560.

### DIFF
--- a/common/cmake/ALPSEnableEigen.cmake
+++ b/common/cmake/ALPSEnableEigen.cmake
@@ -123,7 +123,8 @@ function(add_eigen)
  3. Rerun CMake with option:
      -DALPS_INSTALL_EIGEN=true 
     to request the installation script to attempt to download Eigen3
-    and co-install it with ALPSCore.
+    and co-install it with ALPSCore. (If it still fails, remove your
+    CMakeCache.txt and try again).
  
     In the latter case, you may optionally also set:
      -DALPS_EIGEN_UNPACK_DIR=<path to directory to unpack Eigen> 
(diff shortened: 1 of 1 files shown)


egull:
Thanks. Looks good to me, I think the most useful info is the warning message and instructions to remove CMakeCache.txt if things fail.

PR #577, opened 2018-10-24
Fix is_zero.hpp

wistaria:
No is_float in C++11
Fixed is_zero for float
Add unit test


galexv:
The PR should have been made against the master branch (my bad, the devel branch should be removed altogether to avoid confusion, or it should be synchronized with master).

I have resubmitted this as PR #579, therefore i am closing this PR.

PR #578, opened 2018-10-24
Fix is_zero.hpp (re-submitted PR #577)

galexv:
This is PR #577 from @wistaria, but based on master rather than devel branch

From the original PR:
> No is_float in C++11
Fixed is_zero for float
Add unit test


galexv:
Sorry, this PR is a bit messed up.
Never mind, i will remove this one and resubmit.

PR #579, opened 2018-10-24
Fix is_zero.hpp (resubmitted against the master branch)

galexv:
Fixed is_zero for float
Add unit test

### DIFF
--- a/utilities/include/alps/numeric/is_zero.hpp
+++ b/utilities/include/alps/numeric/is_zero.hpp
@@ -10,6 +10,8 @@
 #define ALPS_NUMERIC_IS_ZERO_HPP
 
 #include <cmath>
+#include <complex>
+#include <limits>
 #include <type_traits>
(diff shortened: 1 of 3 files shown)


galexv:
This is PR #577 by @wistaria  (Synge Todo) resubmitted against the master branch.

utilities/include/alps/numeric/is_zero.hpp
```diff
@@ -10,6 +10,8 @@
 #define ALPS_NUMERIC_IS_ZERO_HPP
 
 #include <cmath>
+#include <complex>
```
galexv: Do we need `#include <complex>` here?
```suggestion
```

wistaria: I think we need <complex> here, since we have a specialized version of is_zero_helper for std::complex<T>.

Issue #482, opened 2018-02-27
[alea] Add tests for `operator==()`


mwallerb:
@galexv's comment on PR #477 

> Sorry for harping on this again :), but how do we know it?
> It would be nice to have a test verifying that `in==out` is `true` iff `in` and `out` are equal, for usual cases and some reasonable corner cases like `count()==0`.

PR #585, opened 2018-11-05
[alea] Add test for operator== and operator!=

mwallerb:
Fixes: #482

### DIFF
--- a/alea/test/compare.cpp
+++ b/alea/test/compare.cpp
@@ -1,0 +1,88 @@
+/*
+ * Copyright (C) 1998-2018 ALPS Collaboration. See COPYRIGHT.TXT
+ * All rights reserved. Use is subject to license terms. See LICENSE.TXT
+ * For use in publications, see ACKNOWLEDGE.TXT
+ */
+#include <alps/alea.hpp>
+#include <alps/alea/util/model.hpp>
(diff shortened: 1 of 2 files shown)


egull:
@galexv this is in response to one of your comments, please have a look at it.

alea/test/compare.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+    void test_one()
+    {
+        Acc acc1(2), acc2(2);
+
+        fill(acc1, 0, 1);
+        EXPECT_NE(acc1.result(), acc2.result());
+
+        acc2 = acc1;
+        EXPECT_EQ(acc1.result(), acc2.result());
```
galexv: We do rely here on the fact that assignment operator makes objects equal.
Or, in other words, "equal objects are, _by definition_, those which are made by an assignment operator".

Markus noted that IEEE FP standard does not even guarantee that two results made from identical accumulators will be strictly numerically equal --- but if they are indeed not equal on some specific platform and this test fails, we would like to know about it.

mwallerb: Good point, thanks for making it explicit.
The test is mainly supposed to take of two corner cases:
 - 0 data points, where the mean does not exist
 - 1 data point, where the variance does not exist

PR #586, opened 2018-11-07
[params] Fix tests for 32-bit OS

wistaria:
Fix tests for 32-bit OS, where sizeof(int) == sizeof(long)

### DIFF
--- a/params/test/dictionary_hdf5.cpp
+++ b/params/test/dictionary_hdf5.cpp
@@ -24,7 +24,7 @@ namespace de=ap::exception;
 using ap::dictionary;
 
 typedef std::vector<bool> boolvec;
-typedef std::vector<int> intvec;
+typedef std::vector<long> longvec;
 typedef std::vector<double> dblvec;
 typedef std::vector<std::string> strvec;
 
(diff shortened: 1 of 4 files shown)


params/test/dictionary_hdf5.cpp
```diff
@@ -24,7 +24,7 @@ namespace de=ap::exception;
 using ap::dictionary;
 
 typedef std::vector<bool> boolvec;
-typedef std::vector<int> intvec;
+typedef std::vector<long> longvec;
```
egull: Confused here by the replacement of int with long. Why replace it rather than test for both? On most platforms the two are not the same. Am I missing something?

egull:
Alex, I'm assigning this to you to look at carefully on Friday. Many of these tests are in your parameters.

params/test/dictionary_hdf5.cpp
wistaria: Sorry for confusion. In the original test, integer is first stored in dict1_ as int. However, after reload from the hdf5 archive, it is stored in dist2_ as long on 32-bit OS, since long and int have the same size. So, dict1_ == dict2_ fails.
Maybe it would be much better to change the logic of operator==() in params, instead of writing such confusing tests...

galexv: I will look at it.

Historical note:
> The idea was, if i remember correctly, that params keep long numbers as long and int numbers as int, and there should be no unexpected overflows, and signed values must compare properly: that's why it has extensive tests. I considered saving all integers as `long` (signed or unsigned), but that complicated treatment of vectors of integers. After some experience with testing and using params, I am now in favor of keeping them simple (all integers are `int_least64_t` and `uint_least64_t`, as these types are mandated by C++11) and performing conversion in vectors by-element.

Issue #275, opened 2017-01-18
Tutorial on GFs is desperately needed.


galexv:
...or at least a bit more doxygen comments in the source code.

galexv:
See PR #591 by @iskakoff

PR #591, opened 2018-11-09
[gf] Add tutorial

iskakoff:
Tutorials on Green's functions

Fixes: #275


tutorials/gf/gf.cpp
```diff
@@ -0,0 +1,87 @@
...
+    for (int j = 0; j < m5.extent(); ++j) {
+      g_view(il(i), ic(j)) = i*10+j;
+    }
+  }
+
+  // Check that we changed the original data
+  for (int i = 0, cnt = 0; i < m4.extent(); ++i) {
+    for (int j = 0; j < m5.extent(); ++j, ++cnt) {
+      if(std::abs(g_view(il(i), ic(j)) - g(iw(0), ii(0), it(1), il(i), ic(j))) > 1e-10 ) {
+        std::cerr<<"Something wrong"<<std::endl;
```
egull: This should never trigger. Propose to throw an exception

tutorials/gf/gf.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+  // create copy of the part of the Green's function object for the first three indices equal to (0,0)
+  alps::gf::greenf<double, itime, legendre, chebyshev> g_copy = g(iw(0), ii(0));
+  std::cout<<"Copy of the part of the Green's function object of the total size "<<g_copy.data().size()<<" has been created"<<std::endl;
+
+  for (int i = 0; i < m4.extent(); ++i) {
+    for (int j = 0; j < m5.extent(); ++j) {
+      // check that it is the same as view
+      if(std::abs(g_view(il(i), ic(j)) - g_copy(it(1), il(i), ic(j))) > 1e-10 ) {
+        std::cerr<<"Something wrong"<<std::endl;
```
egull: same here.

tutorials/gf/gf.cpp
```diff
@@ -0,0 +1,87 @@
...
+      }
+      // change the value
+      g_copy(it(1), il(i), ic(j)) += i*4 + 5 + j;
+      // check that it does not affect original
+      if(std::abs(g_view(il(i), ic(j)) - g_copy(it(1), il(i), ic(j))) < 1e-10 ) {
+        std::cerr<<"Something wrong"<<std::endl;
```
egull: same here.

tutorials/CMakeLists.txt
```diff
@@ -25,3 +25,4 @@ if (ALPSCore_HAS_MPI)
 endif()
 add_subdirectory("mc/ising2_mc")
 add_subdirectory("many_points/process")
+add_subdirectory("gf")
```
egull: See Alex' prior comment about newlines

tutorials/gf/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+project(gf CXX)
+
+set(CMAKE_CXX_STANDARD 11)
+
+# the project relies on the ALPSCore package. If ALPSCore is not found
+# automatically, specify its location using:
+# export ALPSCore_DIR=/location/to/ALPSCORE/
+find_package(ALPSCore 2.2 REQUIRED gf)
+
+# check if Eigen is present:
```
galexv: This check (5 lines) is not needed: Eigen is not used by the tutorial code.
```suggestion

```

iskakoff: Eigen is used by tensors, that are the container for GFs

tutorials/gf/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+# This is a minimal CMake file that illustrates how to use Eigen
```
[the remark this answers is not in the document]
iskakoff: yeah, it was copy-paste

tutorials/gf/gf.cpp
```diff
@@ -0,0 +1,87 @@
...
+      }
+    }
+  }
+
+}
+
+
```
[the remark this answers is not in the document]
iskakoff: Until `fourier.hpp` is not written for the general case, I would consider it as it is not ready for production and therefore we should not advertise it.

tutorials/gf/CMakeLists.txt
iskakoff: But it is the same. GF is almost header only, and it will include the Eigen.
And if ALPSCore does not provide Eigen I suppose it's not gonna work then.

galexv: No, it's not the same:

Situation 1, pertinent to Eigen-in-ALPSCore tutorial: Suppose some future version of ALPSCore will not use Eigen in utilities. I built ALPSCore without Eigen, but my program uses Eigen. I want to know whether i need to request it.

Situation 2, pertinent to GF tutorial: Suppose some future version of ALPSCore will not use Eigen even in GF. I built ALPSCore and want to use GF. Even if ALPSCore does not provide Eigen, i do not need to request it my code does not need it.

So, the difference is "my code uses Eigen directly" vs. "my code uses GF, which happens to use Eigen in its current implementation".

egull:
...and rename #include <alps/gf/mesh/frequency_meches.hpp> where this is still wrong.

PR #594, opened 2018-11-09
Make NDEBUG guard for uninitialized Chebyshev mesh

mwallerb:
(I want to try Inception pull request.)

### DIFF
--- a/gf/include/alps/gf/mesh/chebyshev_mesh.hpp
+++ b/gf/include/alps/gf/mesh/chebyshev_mesh.hpp
@@ -24,9 +24,11 @@ namespace alps {namespace gf {
     std::vector<double> zeros_;
 
     inline void throw_if_empty() const {
+#ifndef NDEBUG
       if (extent() == 0) {
         throw std::runtime_error("chebyshev mesh is empty");
       }
+#endif
     }
 
   public:
@@ -40,9 +42,7 @@ namespace alps {namespace gf {
     }
 
     int operator()(index_type idx) const {
-#ifndef NDEBUG
       throw_if_empty();
-#endif
       return idx();
     }
 


gf/include/alps/gf/mesh/chebyshev_mesh.hpp
```diff
@@ -24,9 +24,11 @@ namespace alps {namespace gf {
     std::vector<double> zeros_;
 
     inline void throw_if_empty() const {
+#ifndef NDEBUG
```
galexv: I have reservations about this. 
This way `throw_if_empty()` is no-op everywhere if `NDEBUG` is set; but it can be argued that `throw_if_empty()` should be active even with `NDEBUG` in the code that is unlikely performance-critical, like `save()` https://github.com/ALPSCore/ALPSCore/blob/74430123f64e5751aa189df24908d69222c796c0/gf/include/alps/gf/mesh/chebyshev_mesh.hpp#L70-L71 , `broadcast()` https://github.com/ALPSCore/ALPSCore/blob/74430123f64e5751aa189df24908d69222c796c0/gf/include/alps/gf/mesh/chebyshev_mesh.hpp#L109-L112 , and possibly `operator !=()` https://github.com/ALPSCore/ALPSCore/blob/74430123f64e5751aa189df24908d69222c796c0/gf/include/alps/gf/mesh/chebyshev_mesh.hpp#L65-L66

I can imagine having conditional `throw_if_empty()`, something like:
```C++
#ifdef NDEBUG
#define LOCAL_DEBUG 0
#else
#define LOCAL_DEBUG 1
#endif
    void throw_if_empty(bool debug=LOCAL_DEBUG) { 
        if (debug && extent() == 0) {
             throw std::runtime_error("chebyshev mesh is empty");
        }
    }; 
#undef LOCAL_DEBUG
```
and calling it as `throw_if_empty()` normally, and `throw_if_empty(true)` if we want it be called even with `NDEBUG` set. Then compiler will likely optimize out the conditional branch and `throw` in case of `NDEBUG` (verified it using compiler explorer, https://godbolt.org/z/mZ-R6z :) )

iskakoff: I think to remove any checks in release mode is the proper way to do stuff.
In other meshes we, e.g., have call of `throw_if_empty` in `operator()` surrounded by `ifndef NDEBUG`.

 In this case I don't see a point to have `throw_if_empty` to be enabled in one part of the code and disabled in other.

PR #595, opened 2018-11-10
[gf] Overhaul of Fourier transforms

iskakoff:
Cleaning up the `fourier.hpp` mess.
I suppose we need to use fft at some point.

Also we need to have tau->omega transform. 
Without this it's sort of pointless.

### DIFF
--- a/gf/include/alps/gf/fourier.hpp
+++ b/gf/include/alps/gf/fourier.hpp
@@ -9,130 +9,118 @@
 namespace alps {
 namespace gf {
 
-inline std::complex<double> f_omega(double wn, double c1, double c2, double c3) {
-  std::complex<double> iwn(0., wn);
-  std::complex<double> iwnsq=iwn*iwn;
-  return c1/iwn + c2/(iwnsq) + c3/(iwn*iwnsq);
-}
+  inline std::complex<double> f_omega(double wn, double c1, double c2, double c3) {
+    std::complex<double> iwn(0., wn);
+    std::complex<double> iwnsq=iwn*iwn;
+    return c1/iwn + c2/(iwnsq) + c3/(iwn*iwnsq);
+  }
 
-inline double f_tau(double tau, double beta, double c1, double c2, double c3) {
-  return -0.5*c1 + (c2*0.25)*(-beta+2.*tau) + (c3*0.25)*(beta*tau-tau*tau);
-}
+  inline double f_tau(double tau, double beta, double c1, double c2, double c3) {
+    return -0.5*c1 + (c2*0.25)*(-beta+2.*tau) + (c3*0.25)*(beta*tau-tau*tau);
+  }
 
-///Fourier transform kernel of the omega -> tau transform
-inline void transform_vector_no_tail(const std::vector<std::complex<double> >&input_data, const std::vector<double> &omega, std::vector<double> &output_data, const std::vector<double> &tau, double beta){
-  for (unsigned int i=0; i<output_data.size(); ++i) {
-    output_data[i] = 0.;
-    for (unsigned int k=0; k<input_data.size(); ++k) {
-      double wt=omega[k]*tau[i];
-      output_data[i]+= 2/beta*(cos(wt)*input_data[k].real()+sin(wt)*input_data[k].imag());
+
+  template<size_t D>
+  inline alps::numerics::tensor<std::complex<double>, D> f_omega(double wn, const alps::numerics::tensor<double, D>& c1, const alps::numerics::tensor<double, D>& c2, const alps::numerics::tensor<double, D>&c3) {
+    std::complex<double> iwn(0., wn);
+    std::complex<double> iwnsq=iwn*iwn;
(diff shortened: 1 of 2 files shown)


gf/include/alps/gf/fourier.hpp
```diff
@@ -9,130 +9,123 @@
...
+    for(int n=0;n<g_omega.mesh1().extent();++n) {
+      in_data(size_t(n)) = g_omega(matsubara_index(size_t(n))).data() - f_omega(g_omega.mesh1().points()[size_t(n)],c1,c2,c3);
     }
-  }
-}
 
+    if(g_omega.mesh1().extent() * g_tau.mesh1().extent() > 5000000 || g_omega.data().size()/g_omega.mesh1().extent() < 100)
+      transform_vector_no_tail_loop(in_data,g_omega.mesh1().points(), g_tau.data(), g_tau.mesh1().points(), g_tau.mesh1().beta());
+    else
+      transform_vector_no_tail_matrix(in_data,g_omega.mesh1().points(), g_tau.data(), g_tau.mesh1().points(), g_tau.mesh1().beta());
```
mwallerb: As discussed with you in person: I think this codepath should be removed,  and the associated function `transform_vector_no_tail_matrix` should be moved to the test.

This switch does not add much benefit, since both codepaths are `O(N^2)`, where `N` is the number of frequencies/time.  Worse, for large `N` we are forced to choose the *slower* codepath.  And, even worse, we know there is a better way than constructing the matrix, which is FFT (or NFFT in the case of an irregular mesh).

This just adds unnecessary complexity, is a medium-time maintenance burden, and is obsolete long-term when we add FFT-backed transforms.

gf/include/alps/gf/fourier.hpp
```diff
@@ -9,130 +9,123 @@
...
-        input_data[n]=g_omega(matsubara_index(n),typename MESH1::index_type(i),typename MESH2::index_type(j))-f_omega(g_omega.mesh1().points()[n],c1,c2,c3);
-      }
-      transform_vector_no_tail(input_data,g_omega.mesh1().points(), output_data, g_tau.mesh1().points(), g_tau.mesh1().beta());
-      for(int t=0;t<g_tau.mesh1().extent();++t){
-        g_tau(itime_index(t),typename MESH1::index_type(i),typename MESH2::index_type(j))=output_data[t]+f_tau(g_tau.mesh1().points()[t],g_tau.mesh1().beta(),c1,c2,c3);
+
+  ///Fourier transform kernel of the omega -> tau transform
+  template<size_t D>
+  inline void transform_vector_no_tail_matrix(const alps::numerics::tensor<std::complex<double>, D>&input_data, const std::vector<double> &omega,
+                                              alps::numerics::tensor<double, D> &output_data, const std::vector<double> &tau, double beta){
```
mwallerb: Propose to move to test (see below)

gf/include/alps/gf/fourier.hpp
```diff
@@ -9,130 +9,123 @@
...
-///Fourier transform kernel of the omega -> tau transform
-inline void transform_vector_no_tail(const std::vector<std::complex<double> >&input_data, const std::vector<double> &omega, std::vector<double> &output_data, const std::vector<double> &tau, double beta){
-  for (unsigned int i=0; i<output_data.size(); ++i) {
-    output_data[i] = 0.;
-    for (unsigned int k=0; k<input_data.size(); ++k) {
-      double wt=omega[k]*tau[i];
-      output_data[i]+= 2/beta*(cos(wt)*input_data[k].real()+sin(wt)*input_data[k].imag());
-    }
+
+  template<size_t D>
```
mwallerb: idem.

gf/include/alps/gf/fourier.hpp
```diff
@@ -9,130 +9,123 @@
...
-    for(int n=0;n<g_omega.mesh1().extent();++n){
-      input_data[n]=g_omega(matsubara_index(n),typename MESH1::index_type(i))-f_omega(g_omega.mesh1().points()[n],c1,c2,c3);
-    }
-    transform_vector_no_tail(input_data,g_omega.mesh1().points(), output_data, g_tau.mesh1().points(), g_tau.mesh1().beta());
-    for(int t=0;t<g_tau.mesh1().extent();++t){
-      g_tau(itime_index(t),typename MESH1::index_type(i))=output_data[t]+f_tau(g_tau.mesh1().points()[t],g_tau.mesh1().beta(),c1,c2,c3);
-    }
+
+  template<size_t D>
+  inline alps::numerics::tensor<double, D> f_tau(double tau, double beta,
```
mwallerb: idem.

gf/include/alps/gf/fourier.hpp
iskakoff: I would like to keep both. The matrix one is way faster than the loop version. But as you mentioned in person, in case of large number of frequencies and imaginary time points
the transformation matrix will be extremely huge.

PR #606, opened 2019-01-19
Fixed hdf5 save/load for vector of params

wistaria:


galexv (approved):
Looks good. Thank you!

PR #608, opened 2019-02-08
Reduce cluttering with  temporary files after running legacy accumulator tests

galexv:


galexv:
This is pretty trivial change; whoever has a moment to review, please do --- I will merge after any of you accepts.

Issue #572, opened 2018-10-19
Improve componentization


galexv:
As discussed with Marcus (@mwallerb), and based on my (@galexv) recent experience with an unrelated project in which I wanted to use the ALPSCore library.

At times, having ALPSCore components depend on multiple other components is inconvenient. For example, IMO `alps::params` provides a decent interface to command-line arguments and parameter files; however, it depends on the archive / HDF5 component which brings possibly unwanted HDF5 dependency. In the same way, one may want to use the MC component, but not want to use HDF5 or params.

It would be good, and is certainly quite feasible, to decrease the interdependence of the components: e.g. by accessing the required functionality between the components via abstract interfaces. In fact, this will bring the ALPSCore components closer to the notion of [software components in the software-engineering sense](https://en.wikipedia.org/wiki/Component-based_software_engineering#Software_component). 

This will also make ALPSCore more appealing to the users of the library.

(@egull This does not fall completely into any existing epic. Shall we create a new one, "componentizaion"?)

mwallerb:
Strong :+1: on this one.

This will bring down compile times as well.

PR #615, opened 2019-04-12
Ability to decouple alps::params from alps::hdf5

galexv:
This PR conditionally decouples `alps::params` from `alps::hdf5`. Specifically, if ALPSCore is built with
`-DALPS_MODULES_DISABLE=hdf5`, the `params` module will be built without archive support. This is useful if all I need from ALPSCore is its `utilities` and `params` modules.

The disadvantage is that the version of ALPSCore build like this is not compatible with the "regular" ALPSCore (e.g., it has different definition of the `alps::params` class).

This PR partially addresses issue #572.

### DIFF
--- a/params/CMakeLists.txt
+++ b/params/CMakeLists.txt
@@ -28,7 +28,9 @@ endif()
 add_this_package(params dict_value dictionary iniparser_interface EXTRA $<TARGET_OBJECTS:libiniparser>)
 
 add_boost()
-add_hdf5()
+if (ALPS_HAVE_ALPS_HDF5)
+    add_hdf5()
+endif()
 
 add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/../common/deps/iniparser" "./iniparser")
 if (ALPS_BUILD_SHARED OR ALPS_BUILD_PIC)
@@ -39,7 +41,12 @@ target_include_directories(${PROJECT_NAME} PRIVATE $<TARGET_PROPERTY:libiniparse
 # This does not work in CMake 3.1:
 # target_sources(${PROJECT_NAME} PRIVATE $<TARGET_OBJECTS:libiniparser>)
 
-add_alps_package(alps-utilities alps-hdf5)
+if (ALPS_HAVE_ALPS_HDF5)
+    add_alps_package(alps-utilities alps-hdf5)
+else()
+    message("NOTE: module params will be built without HDF5 support (alps-hdf5 is disabled)")
+    add_alps_package(alps-utilities)
+endif()
 
 add_testing()
 
--- a/params/include/alps/dictionary.hpp
+++ b/params/include/alps/dictionary.hpp
@@ -79,11 +79,13 @@ namespace alps {
             /// Compare two dictionaries (true if all entries are of the same type and value)
             bool equals(const dictionary& rhs) const;
 
+#ifdef ALPS_HAVE_ALPS_HDF5
             /// Save the dictionary to an archive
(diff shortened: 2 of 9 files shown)


egull:
Markus, this is near and dear to you. Does it help you?

mwallerb:
@galexv Quick question: is with/without HDF really ABI-incompatible?  Usually, adding methods to a class should not break that, or am I wrong?  If so, would making the methods `virtual` help?

galexv:
@mwallerb (I should have said more clearly that it's a semi-measure: i do need ALPSCore's params in a different project and i do not want to carry anything else with it).

I am not sure about binary incompatibility, but certainly [ODR](https://en.cppreference.com/w/cpp/language/definition#One_Definition_Rule) is violated: an application compiled with the regular ALPSCore should not use "lightweighted" ALPSCore libraries, because the app and the lightweight libraries used different definition of the `params` class in their headers.

Could you clarify possible concerns --- in what sense you need the binary compatibility? 

For example, if a user's app needs the lightweight ALPSCore, the user can recompile ALPSCore excluding the unwanted modules and link to the application. The drawback is that there could be a proliferation of local ALPSCore versions each with its own set of compiled modules, which potentially may lead to confusion. That's why this PR is only a partial solution.

mwallerb:
@galexv OK, maybe I typed too fast.  Anyway, perhaps as a tangent, a way to *actually* decouple parameters from HDF5 would be to negotiate it via a purely abstract base class. 

Take a look at the ALEA's [serializer] interface and it's [HDF5 implementation]: it exposes exactly as much of HDF5 as needed for the serialization of arrays to disk.  Crucially though, the HDF5 dependency is broken in compilation and linking, because the connection is only made in the user code.

I think this has a number of advantages to the preprocessor switches: it is less brittle in the case of a mixed setup and it also improves compilation time.  In theory you have to pay a price for dispatch, but since most HDF5 calls involve I/O this should not matter.

I don't know the parameter part of ALPS well, so I have difficulty estimating the effort ... but at least there is some prior work in ALEA that one can rely on.  So here's the proposal:

 1. elevate `alea::serializer` and `alea::deserializer` to common ALPS classes.
 2. specialize `serializer::write()` and `deserializer::read()` for types needed by parameters
 3. add factory methods to `hdf5_archive` class that return `serializer` / `deserializer` instances.  Something to the effect of:

         class hdf5_archive {
              class hdf5_serializer : public serializer { ... };
              class hdf5_deserializer : public deserializer { ... };
              ...
              hdf5_serializer get_serializer();
              hdf5_deserializer get_deserializer() const; // not sure about const
         }

 4. Template `save()` and `load()` methods of parameters over archive type, call the factory method to obtain serializer, pass a reference the underlying ABC to the compiled code.
 5. Profit!

What do you think?

To summarize, I'm totally okay with the current approach if manpower is at a premium, but above would be The Proper Way<sup>TM</sup> to do it IMHO.

[serializer]: https://github.com/ALPSCore/ALPSCore/blob/master/alea/include/alps/alea/core.hpp#L320
[HDF5 implementation]: https://github.com/ALPSCore/ALPSCore/blob/master/alea/include/alps/alea/hdf5.hpp#L18

egull:
I agree with that. There are a bunch of places where interfaces rather than dependencies will help a lot – this is one of them. Alex, we can discuss this on Monday...

galexv:
@mwallerb @egull I agree with this assessment. We can discuss it further on Monday. I'd still think that this PR may be merged as a temporary solution.

PR #616, opened 2019-04-18
Elevate ALEA (`de`)`serializer` interface to common.

mwallerb:
This adds the `alps::serialization` namespace.  The ALEA serialization and deserialization infrastructure is moved from `alps::alea` to this namespace, and ALEA imports those back into the `alps::alea` namespace for backwards compatibility.

The tests for serializer remain in the `alea` package for now, since its the only user.

This is step 1 in decoupling parameters from HDF5.

### DIFF
--- a/alea/include/alps/alea/serialize.hpp
+++ b/alea/include/alps/alea/serialize.hpp
@@ -9,183 +9,18 @@
 
 #include <array>
 
-namespace alps { namespace alea { namespace internal {
 
-/**
- * Allows RAII-type use of groups.
- */
-struct serializer_sentry
-{
-    serializer_sentry(serializer &ser, const std::string &group)
-        : ser_(ser)
-        , group_(group)
-    {
-        if (group != "")
-            ser_.enter(group);
-    }
-
-    ~serializer_sentry()
-    {
-        if (group_ != "")
-            ser_.exit();
-    }
-
-private:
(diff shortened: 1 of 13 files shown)


alea/include/alps/alea/serialize.hpp
```diff
@@ -6,186 +6,65 @@
...
-                    && value.rows() != 1 && value.cols() != 1))
-        serialize(ser, key, plain_matrix_type(value));
+    // Ensure that evaluated expression will be Fortran-contiguous
+    if (!eigen_is_contiguous(value))
+        serialize(ser, key, matrix_type(value));
 
     // Evaluate to matrix or proxy object if already matrix
-    auto temp = value.eval();
-
+    const double *dbl_data = reinterpret_cast<const double *>(value.data());
```
galexv: This `reinterpret_cast` (and another one below) looked scary enough for me to spend some time on cppreference and SE, but i think i got it finally. :) Is it to deal with complex numbers? 
In this case, it could be nice to add a comment here, some time down the road -- no need to delay the merge because of this.

mwallerb: You're absolutely right ... it's to deal with `complex_op` entities, which are essentially 2x2 matrices, laid out in memory like `double[2][2]`.

PR #621, opened 2020-07-28
Tensor update

xinyangd:


utilities/include/alps/numeric/tensors/tensor_base.hpp
```diff
@@ -316,10 +316,12 @@ namespace alps {
         /**
          * Inplace tensor multiplication
          */
-        template<typename S>
-        typename std::enable_if < std::is_same < S, tensorType >::value, tensorType & >::type operator*=(const S& rhs) {
+        template<typename S, typename Ct>
+        typename std::enable_if <
+          std::is_same < S, T >::value || std::is_same < T, std::complex < double>>::value
```
iskakoff, line 321: why? what if T=complex<double> and S=complex<float>?
it is perfectly valid for *= operation, but it will not work in your case. You need to change this enable_if
in a way it properly handles this situation. For this you can use something like "std::is_convertible".

Could you also write a test that will take tensors with different scalar types and call this operation?
something like:
tensor<double, 2> a(N, N);
tensor<float, 2> b(N, N);
tensor<std::complex<double>, 2> c(N, N);
tensor<std::complex<float>, 2> d(N, N);

Initialize them with some values and then you should have something like this:

tensor<double, 2> ab = a * b;
tensor<std::complex<double>, 2> ac = a * c;
tensor<std::complex<float>, 2> bd = b * d;
tensor<std::complex<double>, 2> cd = c * d;

and check that the result of this product is correct.

utilities/test/tensor_test.cpp
```diff
@@ -540,3 +540,33 @@ TEST(TensorTest, Negate) {
...
+  using Matrix = Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>;
+
+  size_t N = 10;
+
+  tensor<double, 3> A(N, N, N);
+  for (size_t i = 0; i < N; ++i) {
+    A(i).matrix() = Matrix::Random(N, N);
+  }
+  const tensor<double, 2> B = A(0);
+  auto X = B.matrix();
```
iskakoff, line 554: replace auto with exact type. it's a test for the method.

utilities/test/tensor_test.cpp
```diff
@@ -540,3 +540,33 @@ TEST(TensorTest, Negate) {
...
+
+  for (int i = 0; i < N; ++i) {
+    for (int j = 0; j < N; ++j) {
+      ASSERT_EQ(X(i, j), A(0, i, j));
+    }
+  }
+
+  auto Dim2 = B.array() * B.array();
```
iskakoff, line 562: same here

utilities/test/tensor_test.cpp
```diff
@@ -540,3 +540,33 @@ TEST(TensorTest, Negate) {
...
+  auto ViewDim2 = A(0).array() * A(0).array();
```
iskakoff, line 563: and here

utilities/test/tensor_test.cpp
```diff
@@ -540,3 +540,33 @@ TEST(TensorTest, Negate) {
...
+  ASSERT_DOUBLE_EQ((Dim2 - ViewDim2).maxCoeff(), 0.);
+
+  auto V1 = A(0, 0).vector();
```
iskakoff, line 566: and also here

utilities/test/tensor_test.cpp
```diff
@@ -540,3 +540,33 @@ TEST(TensorTest, Negate) {
...
+  const tensor<double, 1> C = A(0, 0);
+  auto V2 = C.vector();
```
iskakoff, line 568: and of course here

PR #625, opened 2021-04-30
No boost bind

iskakoff:
Remove deprecated boost::bind


katherlee (approved):
Looks good to me. Just a question: there's no performance advantage of using `std::bind` instead of a lambda capture right?

PR #626, opened 2021-05-06
fix move semantics

iskakoff:


utilities/include/alps/numeric/tensors/tensor_base.hpp
```diff
@@ -207,17 +207,70 @@ namespace alps {
...
+          acc_sizes_ = rhs.acc_sizes();
           storage_ = rhs.storage();
           return *this;
-        };
+        }
+        /// Move assignment
+        tensor_base < T, Dim, Container > & operator=(tensor_base < T, Dim, Container > &&rhs) {
+          shape_ = std::move(rhs.shape());
+          acc_sizes_ = std::move(rhs.acc_sizes());
+          storage_ = std::move(rhs.storage());
```
katherlee, line 224: Question: What happens if `storage_` is a view?

utilities/include/alps/numeric/tensors/tensor_base.hpp
```diff
@@ -207,17 +207,70 @@ namespace alps {
...
+        typename std::enable_if<!std::is_same<Container2, Container>::value, tensor_base < T, Dim, Container > >::type & 
+        operator=(tensor_base < T, Dim, Container2 > &&rhs) {
+          shape_ = rhs.shape();
+          acc_sizes_ = rhs.acc_sizes();
+          reshape(shape_);
+          std::copy(rhs.storage().data(), rhs.storage().data() + rhs.storage().size(), storage_.data());
+          return *this;
+        }
+
+        // assignment for different rhs value type
```
katherlee, line 250: Looks like the following cases can be covered by the ones above by adding the `T2` template parameter?

katherlee (commented):
Looks good overall, just a few questions.

utilities/include/alps/numeric/tensors/tensor_base.hpp
iskakoff: For T != T2 I need to do deep copy instead of move.
The copy constructor indeed can be replaced.

katherlee: But for different containers you'll have to do a copy for move assignments anyways right?

utilities/include/alps/numeric/tensors/tensor_base.hpp
iskakoff: `data_view` stores data in the `view` object, and it has deep copy for move assignment, because we can not move pointers between views.

utilities/include/alps/numeric/tensors/tensor_base.hpp
iskakoff: oh, I see what you mean. yes

iskakoff: ok. done I suppose

katherlee (approved):
LGTM

egull:
Guys, could you please put tests that test the various move scenarios?

common/build/Jenkinsfile.pauli
```diff
@@ -91,30 +91,30 @@ pipeline {
         }
         unstable {
             echo 'DEBUG: Build is unstable'
-            emailext to: 'galexv+jenkins.status@umich.edu',
-                     recipientProviders: [brokenTestsSuspects(), culprits(), requestor()],
-                     subject: 'ALPSCore: Jenkins build is unstable',
-                     attachLog: true,
-                     compressLog: true,
-                     body: "ALPSCore build is unstable: see attached log"
+            //emailext to: 'galexv+jenkins.status@umich.edu',
```
egull, line 94: So what happens now when a test fails? Nothing?

iskakoff: I am not sure it actually worked in the first place. And the last time I checked jenkins, this actually caused an error and test failure because it couldn't send an email.

Also somehow github stopped reading status from our jenkins and I am trying to understand why.

iskakoff:
> Guys, could you please put tests that test the various move scenarios?

there are various scenarios already

egull:
They didn’t discover the problems you had though. Can you find a test that does?

iskakoff:
> They didn’t discover the problems you had though. Can you find a test that does?

theoretically they were supposed to, but yep, I checked they didn't. Now we have proper test for that.

PR #651, opened 2025-10-04
Support for Eigen3 5.0.0

cho-m:
Note that Eigen 5.0.0 is still Eigen3 (EIGEN_WORLD_VERSION=3) but just uses their new version scheme:
- https://gitlab.com/libeigen/eigen/-/releases/5.0.0#versioning

Changes account for:
- Latest Eigen has moved the version information to `Eigen/Version` - https://gitlab.com/libeigen/eigen/-/blob/5.0.0/Eigen/Version
- A build failure I hit:
  ```
  .../accumulators/src/mpi.cpp:56:21: error: use of undeclared identifier 'assert'
     56 |                     assert(buffer.size()>=offset+size && "buffer has sufficient size to accommodate values");
        |                     ^
  ```


cho-m:
Looks like current approach will set 3.5.0 since it always starts with world part.

Maybe just extracting EIGEN_VERSION_STRING will be simpler for 5.0.0+

cho-m:
> * FindEigen has been provided by cmake for some time. It may make sense to use that version instead of maintaining our own and committing the fix. Would that work? Any advantage / disadvantage?

I don't think CMake provides FindEigen. Eigen does ship a Eigen3Config.cmake which does seem to be used by ALPSCore when `EIGEN3_INCLUDE_DIR` is not set.

However, in 5.0.0, Eigen removed `EIGEN3_INCLUDE_DIR` variable and now expects only `Eigen3::Eigen` target usage - https://gitlab.com/libeigen/eigen/-/commit/f2984cd0778dd0a1d7e74216d826eaff2bc6bfab

Also need to account for version checking - https://gitlab.com/libeigen/eigen/-/merge_requests/1989/diffs, e.g.
* `find_package(Eigen3)` would find both 3.4.0, 3.4.1 and 5.0.0
* `find_package(Eigen3 3.4)` would only find 3.4.0
* `find_package(Eigen3 3.4....5)` would find 3.4.1 and 5.0.0

PR #652, opened 2025-10-05
upgrading gtest version and deprecating old cmake versions to suppres…

egull:
…s warnings. Also switching boost to the BoostConfig.cmake file

...merging this since we currently have nobody else to review this.


egull:
Looks good. Unrelated MPI error in the cloud which we really ought to fix.

PR #654, opened 2025-10-05
Revert 652 maintenance

egull:


egull:
...that pull request was too early. Reverting it.

PR #659, opened 2026-05-05
Fix dangling tensor_view in tensor_test helper

matthiastroyer:
## Problem
The `make_tensor` helper function in `utilities/test/tensor_test.cpp` was returning a non-owning `tensor_view<double, 2>` that pointed to a local `tensor<double, 3>`. When the function returned, the local tensor was destroyed, leaving the view with a dangling pointer to freed memory.

This caused `TensorTest.TestMoveAssignments2` to fail with undefined behavior when reading from the dangling view.

## Root Cause
`tensor_view` is a non-owning view type that only holds pointers to data. Even with `std::move()`, the view still references destroyed memory because the actual tensor (which owned the data) went out of scope.

## Solution
Changed `make_tensor()` to return an owning `tensor<double, 2>` instead of a non-owning view. The owning tensor carries its data with it via move semantics, eliminating the lifetime issue.

## Testing
- All 107 tests now pass (previously 106/107)
- No other tests affected by this change
- Validation: `ctest --output-on-failure -j4` ✅

### DIFF
--- a/utilities/test/tensor_test.cpp
+++ b/utilities/test/tensor_test.cpp
@@ -124,8 +124,8 @@ TEST(TensorTest, TestMoveAssignments2) {
   };
   tensor<double, 2> T1(N, N);
   tensor<double, 2> T2(N, N);
-  T1 = move_tensor(N, 0, 2);
-  T2 = move_tensor(N, 1, 3);
+  T1 = move_tensor(T, N, 0, 2);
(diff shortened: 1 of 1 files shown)


iskakoff (approved):
With correct fix to the dangling memory, this PR can now be merged.
I checked with valgrind, this test no longer accesses already freed memory.

PR #667, opened 2026-09-23
Include <algorithm> where standard algorithms are used (fixes GCC 15 build)

egull:
## Problem

Building ALPSCore with GCC 15 (libstdc++; MacPorts `g++-mp-15`) fails:

```
params/include/alps/params/params_impl.hpp:67:42: error: no matching function for call to 'find(sit_type&, std::__cxx11::basic_string<char>::const_iterator, char)'
```

## Cause

`params_impl.hpp` calls unqualified `find(it1, in.end(), ',')` but never includes `<algorithm>`. With libc++ (Clang) the header comes in transitively. With current libstdc++ it does not, and ADL only finds unrelated `std::find` overloads. Adding `-include algorithm` to `CMAKE_CXX_FLAGS` makes the build succeed, which confirms the cause.

## Fix

- `params_impl.hpp`: add `#include <algorithm>` and call `std::find`.
- I searched the whole tree (all modules, headers, sources and tests; `common/deps` excluded) for `<algorithm>` functions used in files that do not include `<algorithm>` themselves. I added the include to those files: `std::min`/`max`, `copy`, `fill`/`fill_n`, `equal`, `sort`, `transform`, `max_element` and `upper_bound` in accumulators, alea, gf, hdf5, params and utilities. All of these calls were already `std::`-qualified, so this change only adds includes. No other code changes.

## Verification

- **GCC 15.x** (`g++-mp-15`, libstdc++), `-DALPS_CXX_STD=c++17 -DENABLE_MPI=OFF -DTesting=ON`, **without** the `-include algorithm` workaround: builds cleanly. ctest: 106/107 pass. The one failure is `four_index_gf_pn_test`, which crashes intermittently with a bus error. This crash comes from a bug that already exists and is not caused by this PR. In that test, `gf2(omega2,...)` with `omega2=-4` on a `POSITIVE_NEGATIVE` Matsubara mesh reaches `gf_base::value()` → `data_(inds()...)`. That path passes the raw negative index to the tensor without adding the mesh offset, so the write lands 1680 bytes before the data buffer. Whether it crashes depends on heap layout. The same test passed 40/40 times in a build of unmodified master that used `-include algorithm`, but the out-of-bounds write happens there too.
  - Note: on macOS 26 with GCC 15, the bundled gtest (`gtest-port.cc` → `<mach/message.h>`) does not compile against the MacOSX26 SDK. I set `-DCMAKE_OSX_SYSROOT=.../MacOSX27.sdk` to work around this. It is unrelated to this change.
- **AppleClang 21** (libc++), `-DENABLE_MPI=ON` (MacPorts Open MPI, openmpi-clang20), `-DTesting=ON`: builds cleanly. ctest: 135/135 pass.


egull:
Approving and merging myself since I'm currently the only one working on it...

